PR #12, opened 2020-06-25
Square/tests

jacobrobertbaca:
# Description of change
wrote tests for tap-square

# Manual QA steps
 - ran tap-tester locally
 
# Risks
 - low
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_automatic_fields.py
+++ b/tests/test_automatic_fields.py
@@ -51,9 +51,8 @@ class TestAutomaticFields(TestSquareBase):
 
         # ensure data exists for sync streams and set expectations
         expected_records = defaultdict(list)
-        start_date = "2020-03-01T00:00:00Z"
         for stream in self.testable_streams():
-            existing_objects = self.client.get_all(stream, start_date)
+            existing_objects = self.client.get_all(stream, self.START_DATE)
             if existing_objects:
                 print("Data exists for stream: {}".format(stream))
                 for obj in existing_objects:
--- a/tests/test_automatic_fields.py
+++ b/tests/test_automatic_fields.py
@@ -1,0 +1,170 @@
+import os
+import unittest
+
+from collections import defaultdict
+from functools import reduce
+
+from singer import metadata
+import tap_tester.connections as connections
+import tap_tester.menagerie   as menagerie
+import tap_tester.runner      as runner
+
+from base import TestSquareBase
+from test_client import TestClient
+
+
+class TestAutomaticFields(TestSquareBase):
+    """Test that with no fields selected for a stream automatic fields are still replicated"""
+
+    def name(self):
+        return "tap_tester_square_automatic_fields"
+
+
+    def testable_streams(self):
+        return self.expected_streams().difference(set())
+
+    def expected_automatic_fields(self):
+
+        auto_fields = {}
+        for k, v in self.expected_metadata().items():
+            auto_fields[k] = v.get(self.PRIMARY_KEYS, set()) | v.get(self.REPLICATION_KEYS, set())
+        return auto_fields
+
+    @classmethod
+    def setUpClass(cls):
+        print("\n\nTEST SETUP\n")
+        cls.client = TestClient()
+
+    @classmethod
+    def tearDownClass(cls):
+        print("\n\nTEST TEARDOWN\n\n")
+
+    def test_run(self):
+        """
+        Verify that for each stream you can get data when no fields are selected
+        and only the automatic fields are replicated.
+        """
+
+        print("\n\nRUNNING {}\n\n".format(self.name()))
+
+        # TODO ensure multiple pages of data exist
+
+        # ensure data exists for sync streams and set expectations
+        expected_records = defaultdict(list)
+        start_date = "2020-03-01T00:00:00Z"
+        for stream in self.testable_streams():
+            existing_objects = self.client.get_all(stream, start_date)
+            if existing_objects:
+                print("Data exists for stream: {}".format(stream))
+                for obj in existing_objects:
+                    expected_records[stream].append(
+                        {field: obj.get(field)
+                         for field in self.expected_automatic_fields().get(stream)}
+                    )
+            else:
+                print("Data does not exist for stream: {}".format(stream))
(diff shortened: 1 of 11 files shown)


tests/test_automatic_fields.py
```diff
@@ -0,0 +1,158 @@
...
+        catalogs = menagerie.get_catalogs(conn_id)
+
+        # Ensure our selection worked
+        for cat in found_catalogs:
+            catalog_entry = menagerie.get_annotated_schema(conn_id, cat['stream_id'])
+            # Verify all streams are selected
+            selected = catalog_entry.get('annotated-schema').get('selected')
+            print("Validating selection on {}: {}".format(cat['stream_name'], selected))
+            self.assertTrue(selected, msg="Stream not selected by default")
+            # TODO check selection for fields
```
kspeer825: I have this almost-done on a branch

kspeer825: getting None when I expect True so not sure if that is correct or not

tests/test_bookmarks.py
```diff
@@ -0,0 +1,145 @@
...
+        return_value = {
+            'start_date' : '2020-06-01T00:00:00Z',
+            'sandbox' : 'true'
+        }
+
+        if original:
+            return return_value
+
+        return_value['start_date'] = self.START_DATE
+        return return_value
```
kspeer825, line 23: I saw this defined in base.py I would drop it from here

tests/test_client.py
```diff
@@ -0,0 +1,175 @@
...
+        elif stream == 'discounts':
+            return [obj for page, _ in self.get_catalog('DISCOUNT', start_date, None) for obj in page]
+        elif stream == 'taxes':
+            return [obj for page, _ in self.get_catalog('TAX', start_date, None) for obj in page]
+        elif stream == 'employees':
+            return [obj for page, _ in self.get_employees(None) for obj in page]
+        elif stream == 'locations':
+            return [obj for page, _ in self.get_locations() for obj in page]
+        else:
+            raise NotImplementedError
```
kspeer825, line 59: I don't quite understand what is happening here.

tests/test_bookmarks.py
```diff
@@ -0,0 +1,145 @@
...
+            self, conn_id, self.expected_streams(), self.expected_primary_keys())
+        return sync_record_count
+
+    def setUp(self):
+        pass
+
+    def tearDown(self):
+        pass
+
+    def test_run(self):
```
kspeer825, line 46: Needs to look more like Trello with multiple syncs and updates/creates in between those syncs.

tests/test_pagination.py
```diff
@@ -0,0 +1,111 @@
...
+        For EACH stream add enough data that you surpass the limit of a single
+        fetch of data.  For instance if you have a limit of 250 records ensure
+        that 251 (or more) records have been posted for that stream.
+        """
+        print("\n\nRUNNING {}\n\n".format(self.name()))
+
+        # Ensure tested streams have a record count which exceeds the API LIMIT
+        for stream in self.testable_streams():
+            existing_objects = self.client.get_all(stream, self.START_DATE)
+            if len(existing_objects) <= 1000:
```
kspeer825: Should be data driven use var at top of test

tests/test_pagination.py
```diff
@@ -0,0 +1,111 @@
...
+                self.assertGreater(record_count_by_stream.get(stream, -1), self.API_LIMIT,
+                                   msg="We didn't guarantee pagination. The number of records should exceed the api limit.")
+
+                data = synced_records.get(stream, [])
+                record_messages_keys = [set(row['data'].keys()) for row in data['messages']]
+
+                for actual_keys in record_messages_keys:
+
+                    # Verify that the automatic fields are sent to the target for paginated streams
+                    self.assertEqual(self.expected_automatic_fields().get(stream) - actual_keys,
```
kspeer825: This should be all fields not automatic

tests/test_start_date.py
```diff
@@ -0,0 +1,460 @@
...
+                    )
+
+                else:
+                    raise Exception("Expectations are set incorrectly. {} cannot have a "
+                                    "replication method of {}".format(stream, replication_type))
+
+        # TODO Remove when test complete
+        print("\n\n\tTOOD's PRESENT | The test is incomplete\n\n")
+
+class TestSquareStartDateIncremental(TestSquareBase):
```
kspeer825: Don't want to typically break this up. Might be worth it though. Only did for adroll because of limitations with creating data for incremental streams.

tests/test_bookmarks.py
luandy64: This test needed a different start date and setting it on the TestCase didn't seem to do anything so we decided to override this function

tests/test_client.py
luandy64: For the stream you pass in, call the `get` function from the tap. The `get`s returns 1 or more pages, so we loop over every page and grab every record

tests/test_pagination.py
luandy64: Can you clarify? We grab all records that we can, and check if we have enough (over 1000 records per stream).

Do you want new records every time?

tests/test_bookmarks.py
kspeer825: Ah gotcha, I handled this differently in other tap but I don't think it is necessarily a better way. This should be fine I think.

tests/test_client.py
kspeer825: Interesting. In adroll the test client defines it's own GET function. I am not sure if separating them adds value or not though. My initial reaction to this is it is bad that we are using the method that we are testing to set the expectations. But I do not know if rewriting the function in a different way is any better.

tests/test_pagination.py
kspeer825: oh no just that you defined API_LIMIT at the top of the file and then didn't use it here

tests/test_start_date.py
luandy64: Merged 👍

kspeer825 (approved):

PR #14, opened 2020-07-07
Testing/revise existing tests

kspeer825:
# Description of change
updated tests
schema issue addressed

# Manual QA steps
 - ran tests locally, verified target output
 
# Risks
 - low, schema change to `items`
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_square/schemas/items.json
+++ b/tap_square/schemas/items.json
@@ -1,48 +1,6 @@
 {
-  "type": [
-    "null",
-    "object"
-  ],
   "properties": {
     "id": {
-      "type": [
-        "null",
-        "string"
-      ]
-    },
-    "is_deleted": {
-      "type": [
-        "null",
-        "boolean"
-      ]
-    },
-    "item_data": {
-      "type": [
-        "null",
-        "object"
-      ],
-      "properties": {
-        "name": {
-          "type": [
-            "null",
-            "string"
-          ]
-        },
-        "product_type": {
-          "type": [
-            "null",
-            "string"
-          ]
-        }
-      }
-    },
-    "present_at_all_locations": {
-      "type": [
-        "null",
-        "boolean"
-      ]
-    },
-    "type": {
       "type": [
         "null",
         "string"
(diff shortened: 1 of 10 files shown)


tap_square/schemas/items.json
```diff
@@ -1,45 +1,229 @@
...
         "string"
       ]
     },
-    "is_deleted": {
+    "updated_at": {
       "type": [
         "null",
-        "boolean"
+        "string"
       ]
```
dmosorast: Assuming this will need to be a datetime? Not sure when we feel comfortable combing over the data types. That's something I did last week for adroll.

dmosorast: I'm sure there could be more, just saw that one jump out at me. Going to ignore others as I review.

tests/test_automatic_fields.py
```diff
@@ -130,8 +130,25 @@ def test_run(self):
...
+                        print("\n==== DATA DISCREPANCY ====\n")
+                        print("Expected: {}\n".format(actual_record))
+                        e_record = [record for record in expected_records.get(stream)
+                                    if actual_record.get('eid') == record.get('eid')]
+                        print("Actual: {}\n".format(e_record))
+                        for key in schema_keys:
+                            e_val = e_record[0].get(key)
+                            val = actual_record.get(key)
+                            if e_val != val:
+                                print("\nDISCREPANCEY | KEY {}: ACTUAL: {} EXPECTED {}".format(key, val, e_val))
```
dmosorast, line 157: With Adroll, this caught our missing datetime conversions, will this catch them here?

tests/test_bookmarks.py
```diff
@@ -56,6 +68,18 @@ def test_run(self):
         For EACH stream that is incrementally replicated there are multiple rows of data with
             different values for the replication key
         """
+        # Ensure tested streams have a record count which exceeds the API LIMIT
+        expected_records_1 = {x: [] for x in self.expected_streams()}
+        for stream in self.testable_streams():
+            existing_objects = self.client.get_all(stream, self.START_DATE)
+
+            if len(existing_objects) == 0:
+                # TODO change from 'failure you need data' to 'create data then so we can test'
```
dmosorast, line 77: Neither of these messages appear here 😅 

Could the assertion message include the stream name?

tests/test_bookmarks.py
```diff
@@ -91,13 +113,39 @@ def test_run(self):
...
+        for stream in self.expected_full_table_streams():
+            for record in expected_records_1.get(stream, []):
+                if record.get('id') == updated_records[stream].get('id'):
+                    continue  # do not add the orginal version of the updated record
+                expected_records_2[stream].append(record)
+
+        # ensure validity of expected_records_2
+        for stream in self.testable_streams():
+            if stream in self.expected_incremental_streams():
+                assert len(expected_records_2.get(stream)) == 2, "Expectations are invalid for full table {}".format(stream)
```
dmosorast: Are these messages switched?

dmosorast (changes requested):
The structure looks good. Seems like it's moving towards standardized. Just some messaging, and schema verification questions (that may need changes as a result). Requesting changes!

tap_square/schemas/items.json
kspeer825: Interesting, this seems like a bug in the infer-schema tool to me. I think that should be a datetime.

tests/test_automatic_fields.py
kspeer825: It should catch any discrepancies between the json in our schema file, and what is returned by square's api. If there is a missing field, or an incorrect data type it should be caught. I also put these assertions in the bookmarks test as well. The tap was built using the square client so there isn't as much independence between the tap's client and the test_client compared to Adroll. But having these assertions in the bookmarks test where we set additional expectations based off a POST and a PUT should ensure more independence between EXPECTED and ACTUAL.

tests/test_bookmarks.py
kspeer825: Added stream name. Creates aren't working for all streams yet and I haven't hit this point of missing data. Once one of those things happen I'll put this logic in.

tests/test_bookmarks.py
kspeer825: Fixed

tests/test_automatic_fields.py
kspeer825: This way of comparing schema isn't perfect though. We are not able to catch the case where the api returns a string but we want to type it as a date-time. Which I am not sure if that is something we should even be doing?

dmosorast (approved):

PR #15, opened 2020-07-07
Testing/add all fields test

kspeer825:
# Description of change
Added `all_fields_test.py` to test suite.

# Manual QA steps
 - Ran test locally, reviewed code and removed bad copy pasta from adroll test
 
# Risks
 - None, test change
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/client_tester.py
+++ b/tests/client_tester.py
@@ -1,0 +1,68 @@
+"""
+This is used for testing basic functionality of the test client.
+To run change the desired flags below and use the following command from the tap-tester repo:
+  'python ../tap-adroll/tests/client_tester.py'
+"""
+from test_client import TestClient
+
+
+##########################################################################
+# Testing the TestCLient
+##########################################################################
+if __name__ == "__main__":
+    client = TestClient()
+    START_DATE = '2020-06-24T00:00:00Z'
+
+    # CHANGE FLAGS HERE TO TEST SPECIFIC FUNCTION TYPES
+    test_creates = False
+    test_updates = False
+    test_gets = True
+    test_deletes = False
+
+    # CHANGE FLAG TO PRINT ALL OBJECTS THAT FUNCTIONS INTERACT WITH
+    print_objects = True
+
+    objects_to_test = [ # CHANGE TO TEST DESIRED STREAMS 
+        'items',  # GET - DONE | CREATE - DONE | UPDATE - 
+        'categories',  # GET - DONE | CREATE - DONE | UPDATE - 
+        'discounts',  # GET - DONE | CREATE - DONE | UPDATE - 
+        'taxes',  # GET - DONE | CREATE - DONE | UPDATE - 
+        # 'employees',  # GET - DONE | CREATE -  | UPDATE - 
+        'locations',  # GET - DONE | CREATE - DONE | UPDATE - 
+    ]
+
+    print("********** Testing basic functions of test client **********")
+    if test_gets:
+        for obj in objects_to_test:
+            print("Testing GET (all): {}".format(obj))
+            import pdb; pdb.set_trace() # UNCOMMENT TO RUN 'INTERACTIVELY'
+            existing_obj = client.get_all(obj, START_DATE)
+            if existing_obj:
+                print("SUCCESS")
+                if print_objects:
+                    print("{} Records".format(len(existing_obj)))
+                    # print(existing_obj)
+                continue
+            print("FAILED")
+    if test_creates:
+        for obj in objects_to_test:
+            print("Testing CREATE: {}".format(obj))
+            # import pdb; pdb.set_trace() # UNCOMMENT TO RUN 'INTERACTIVELY'
+            created_obj = client.create(obj)
+            if created_obj:
+                print("SUCCESS")
+                if print_objects:
+                    print(created_obj)
+                continue
+            print("FAILED")
+    if test_updates:
+        for obj in objects_to_test:
+            print("Testing UPDATE: {}".format(obj))
+            # import pdb; pdb.set_trace() # UNCOMMENT TO RUN 'INTERACTIVELY'
+            updated_obj = client.update(obj)
+            if updated_obj:
+                print("SUCCESS")
+                if print_objects:
+                    print(updated_obj)
+                continue
+            print("FAILED")
--- a/tests/test_all_fields.py
+++ b/tests/test_all_fields.py
@@ -1,7 +1,6 @@
 import os
 import unittest
 
-from functools import reduce
 from singer import metadata
 import tap_tester.connections as connections
 import tap_tester.menagerie   as menagerie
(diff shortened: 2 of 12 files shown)


tests/client_tester.py
```diff
@@ -0,0 +1,68 @@
...
+To run change the desired flags below and use the following command from the tap-tester repo:
+  'python ../tap-adroll/tests/client_tester.py'
+"""
+from test_client import TestClient
+
+
+##########################################################################
+# Testing the TestCLient
+##########################################################################
+if __name__ == "__main__":
```
asaf-erlich, line 12: Why is this not a unit test case class? Why is it a script like this?

tests/client_tester.py
```diff
@@ -0,0 +1,68 @@
...
+        'taxes',  # GET - DONE | CREATE - DONE | UPDATE - 
+        # 'employees',  # GET - DONE | CREATE -  | UPDATE - 
+        'locations',  # GET - DONE | CREATE - DONE | UPDATE - 
+    ]
+
+    print("********** Testing basic functions of test client **********")
+    if test_gets:
+        for obj in objects_to_test:
+            print("Testing GET (all): {}".format(obj))
+            import pdb; pdb.set_trace() # UNCOMMENT TO RUN 'INTERACTIVELY'
```
asaf-erlich, line 38: Why is this trace still here? Is it necessary?

tests/client_tester.py
```diff
@@ -0,0 +1,68 @@
...
+            existing_obj = client.get_all(obj, START_DATE)
+            if existing_obj:
+                print("SUCCESS")
+                if print_objects:
+                    print("{} Records".format(len(existing_obj)))
+                    # print(existing_obj)
+                continue
+            print("FAILED")
```
asaf-erlich, line 46: Why are you printing FAILED here?

tests/client_tester.py
kspeer825: This is just a basic script I threw together to aid in dev/debug of the test_client. It helps me do sanity checks and verify root cause for errors in posts and puts, without having to run an entire test

tests/client_tester.py
```diff
@@ -0,0 +1,68 @@
...
+        for obj in objects_to_test:
+            print("Testing CREATE: {}".format(obj))
+            # import pdb; pdb.set_trace() # UNCOMMENT TO RUN 'INTERACTIVELY'
+            created_obj = client.create(obj)
+            if created_obj:
+                print("SUCCESS")
+                if print_objects:
+                    print(created_obj)
+                continue
+            print("FAILED")
```
asaf-erlich, line 57: Same question

tests/client_tester.py
kspeer825: If this statement is executed it tells me that nothing was returned by that request

tests/client_tester.py
```diff
@@ -0,0 +1,68 @@
...
+        for obj in objects_to_test:
+            print("Testing UPDATE: {}".format(obj))
+            # import pdb; pdb.set_trace() # UNCOMMENT TO RUN 'INTERACTIVELY'
+            updated_obj = client.update(obj)
+            if updated_obj:
+                print("SUCCESS")
+                if print_objects:
+                    print(updated_obj)
+                continue
+            print("FAILED")
```
asaf-erlich, line 68: Same

tests/test_all_fields.py
```diff
@@ -0,0 +1,187 @@
...
+        expected_records = {x: [] for x in self.expected_streams()} # ids by stream
+        for stream in self.testable_streams():
+            existing_objects = self.client.get_all(stream, self.START_DATE)
+            if existing_objects:
+                print("Data exists for stream: {}".format(stream))
+                for obj in existing_objects:
+                    expected_records[stream].append(obj)
+            else:
+               print("Data does not exist for stream: {}".format(stream))
+               assert None, "more test functinality needed"
```
asaf-erlich, line 54: Is this asserting that None is not equal to a string? Shouldn't it be a fail("message")?

kspeer825: This is a placeholder, too early to implement a create for all testable streams in this test. And it is not necessary since data exists for all streams under test. By the end each test will have methods to ensure that if data does not exist, it is created prior to the test execution. That way we ensure each test is independent of existing data and of other tests.

tests/test_all_fields.py
```diff
@@ -0,0 +1,187 @@
...
+
+        # Verify tap exit codes
+        exit_status = menagerie.get_exit_status(conn_id, sync_job_name)
+        menagerie.verify_sync_exit_status(self, exit_status, sync_job_name)
+
+        # read target output
+        first_record_count_by_stream = runner.examine_target_output_file(self, conn_id,
+                                                                         self.expected_streams(),
+                                                                         self.expected_primary_keys())
+        replicated_row_count =  reduce(lambda accum,c : accum + c, first_record_count_by_stream.values())
```
asaf-erlich: Instead of reduce can you just use sum here?

kspeer825: Good point. Not sure why reduce was used, I think sum would be more clear.

tests/test_all_fields.py
kspeer825: Oh I misread your comment no, it is asserting None so that the test fails. And then the output is `AssertionError: more test functionality needed`

asaf-erlich: Right, I'm saying https://docs.python.org/3/library/unittest.html#unittest.TestCase.fail is more clear than `assert None, message`.

asaf-erlich: Not a big deal

tests/test_all_fields.py
```diff
@@ -0,0 +1,186 @@
...
+        # Verify target has records for all synced streams
+        for stream, count in first_record_count_by_stream.items():
+            assert stream in self.expected_streams()
+            self.assertGreater(count, 0, msg="failed to replicate any data for: {}".format(stream))
+        print("total replicated row count: {}".format(replicated_row_count))
+
+        # Test by Stream
+        for stream in self.testable_streams():
+            with self.subTest(stream=stream):
+                data = synced_records.get(stream)
```
asaf-erlich, line 124: The test looks fine, but I might refactor the stuff inside the for loop into its own method

asaf-erlich (approved):
LGTM with minor comment, you decide whether the refactor is worth it

tests/test_all_fields.py
kspeer825: What would be the benefit of doing this?

asaf-erlich: The benefits of refactoring in general are to make the code more readable. Having one gigantic test method makes the code hard to read. It's hard to parse and grok what's going on because it's doing so much and checking so many things.

PR #16, opened 2020-07-10
Testing/employees

kspeer825:
# Description of change
Add creates and updates for `employees` stream.
Add `employees` to as many tests as possible

# Manual QA steps
 - Run methods individually and verify functionality in ui
 
# Risks
 - None, Test client and test changes
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -219,12 +219,10 @@ class TestSquareBase(unittest.TestCase):
             for record in expected_records:
                 for key, value in record.items():
                     self.align_date_type(record, key, value)
-                    self.sort_array_type(record, key, value)
             return
 
         for key, value in expected_records.items(): # Modify a single record
             self.align_date_type(expected_records, key, value)
-            self.sort_array_type(expected_records, key, value)
 
     def align_date_type(self, record, key, value):
         """datetime values must conform to ISO-8601 or they will be rejected by the gate"""
@@ -233,13 +231,18 @@ class TestSquareBase(unittest.TestCase):
             iso_date = dt.strftime(raw_date,  "%Y-%m-%dT%H:%M:%S.%fZ")
             record[key] = iso_date
 
-    def sort_array_type(self, record, key, value):
-        try:
-            if isinstance(value, list) and value and key in ['ads']:
-                if isinstance(value[0], dict) and "id" in value[0].keys():
-                    record[key] = sorted(value, key=lambda x: x['id'])
-                else:
-                    record[key] = sorted(value)
-        except Exception as ex:
-            print("Could not sort array at key: {}, value: {}".format(key, value))
-            raise
+    # TODO REMOVE or UNCOMMENT once we determine if this is needed
+    # def sort_array_type(self, record, key, value):
+    #     """
+    #     List values are returned by square as unordered arrays.
+    #     In order to accurately compare expected and actual records, we must sort all lists.
+    #     """
+    #     try:
+    #         if isinstance(value, list) and value: #  and key in ['ads']:
+    #             if isinstance(value[0], dict) and "id" in value[0].keys():
+    #                 record[key] = sorted(value, key=lambda x: x['id'])
+    #             else:
+    #                 record[key] = sorted(value)
+    #     except Exception as ex:
+    #         print("Could not sort array at key: {}, value: {}".format(key, value))
+    #         raise
--- a/tests/base.py
+++ b/tests/base.py
@@ -222,10 +222,12 @@ class TestSquareBase(unittest.TestCase):
             for record in expected_records:
                 for key, value in record.items():
                     self.align_date_type(record, key, value)
+                    self.sort_array_type(record, key, value)
             return
 
         for key, value in expected_records.items(): # Modify a single record
             self.align_date_type(expected_records, key, value)
+            self.sort_array_type(expected_records, key, value)
 
     def align_date_type(self, record, key, value):
         """datetime values must conform to ISO-8601 or they will be rejected by the gate"""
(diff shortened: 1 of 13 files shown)


tests/base.py
```diff
@@ -43,13 +43,14 @@ def setUpClass(cls):
         print("\n\nTEST SETUP\n")
         cls.client = TestClient()
 
-    @classmethod
-    def tearDownClass(cls):
-        print("\n\nTEST TEARDOWN\n\n")
+    # @classmethod
+    # def tearDownClass(cls):
+    #     print("\n\nTEST TEARDOWN\n\n")
```
asaf-erlich: Can you delete this code instead of leaving it commented out

tests/base.py
```diff
@@ -222,14 +223,27 @@ def modify_expected_datatypes(self, expected_records):
...
+            self.sort_array_type(expected_records, key, value)
 
     def align_date_type(self, record, key, value):
         """datetime values must conform to ISO-8601 or they will be rejected by the gate"""
         if isinstance(value, str) and key in ['updated_at']:
             raw_date = self.parse_date(value)
             iso_date = dt.strftime(raw_date,  "%Y-%m-%dT%H:%M:%S.%fZ")
             record[key] = iso_date
+
+    def sort_array_type(self, record, key, value):
```
asaf-erlich: What is the point of this method written in this way? Is this saying only sort if the type is list / array?

tests/base.py
```diff
@@ -222,14 +223,27 @@ def modify_expected_datatypes(self, expected_records):
...
+        try:
+            if isinstance(value, list) and value and key in ['ads']:
```
asaf-erlich: Why are you hardcoding 'ads' and 'id' in this method? Can those be inferred from the schema?

asaf-erlich: I'm a bit confused. Is it asking if both value and key is equal to the word 'ads'? Why write `in ['ads']`?

asaf-erlich: Oh sorry no it's saying "value" is truthy (not empty, not None), but then `key in ['ads']` is the same as `key == 'ads'` right? Why not do that?

tests/client_tester.py
```diff
@@ -15,54 +15,55 @@
...
-        'discounts',  # GET - DONE | CREATE - DONE | UPDATE - 
-        'taxes',  # GET - DONE | CREATE - DONE | UPDATE - 
-        # 'employees',  # GET - DONE | CREATE -  | UPDATE - 
-        'locations',  # GET - DONE | CREATE - DONE | UPDATE - 
+        # 'items',  # GET - DONE | CREATE - DONE | UPDATE - DONE
+        # 'categories',  # GET - DONE | CREATE - DONE | UPDATE - DONE
+        # 'discounts',  # GET - DONE | CREATE - DONE | UPDATE - DONE
+        # 'taxes',  # GET - DONE | CREATE - DONE | UPDATE - DONE
+        'employees',  # GET - DONE | CREATE -  | UPDATE - 
+        # 'locations',  # GET - DONE | CREATE - DONE | UPDATE - DONE
```
asaf-erlich, line 31: Why comment the other things?

tests/test_bookmarks.py
```diff
@@ -18,10 +18,14 @@ def name(self):
...
             {  # STREAMS NOT CURRENTY TESTABLE
-                'employees',
+                'employees', # Requires production environment to create records
                 'locations'
             }
         )
+    @classmethod
+    def tearDownClass(cls):
+        print("\n\nTEST TEARDOWN\n\n")
+        # TODO add delete for locations if possible
```
asaf-erlich, line 28: Can we just add the teardown later when we need it?

tests/base.py
kspeer825: Yes we are sorting arrays in order to make a valid comparison between expected and actual records. It is not always necessary but in this case square returns fields that are arrays as unordered lists. Since we are passing in the whole record set, we are checking the type.

tests/client_tester.py
kspeer825: I was just testing employees last. This isn't executed in circle, so the state of this list isn't too important.

tests/test_bookmarks.py
kspeer825: We can, I like having the print statement because it helps separate the test output in circle. But it definitely isn't needed.

tests/base.py
asaf-erlich: Ok, that makes sense. Can you write a docstring explaining that?

tests/client_tester.py
asaf-erlich: Ok

tests/test_bookmarks.py
asaf-erlich: Ok fair enough that's fine.

tests/base.py
```diff
@@ -222,14 +223,27 @@ def modify_expected_datatypes(self, expected_records):
...
+                if isinstance(value[0], dict) and "id" in value[0].keys():
```
asaf-erlich: You sure you're only supposed to this for the first value? And not for every value? Cause in the next line you sort the value for that key. And if id was not in a later value (which I realize doesn't make much sense but could be possible) it might fail.

tests/base.py
kspeer825: This was a bad copy pasta from adroll. `ads` was a field that was never sorted. At one point I had records that required sorting but based on the last few test runs with the new app/data-set we do not have that issue anymore. I am going to remove this for now and put it back once we end up with records that have list values.

tests/base.py
kspeer825: Done

tests/base.py
kspeer825: Done

tests/base.py
kspeer825: Commented the method and put a TODO to use or remove

tests/base.py
kspeer825: That's a good point, I didn't write this one so I am not sure what the reason was for looking at the first value only. Once I have data for a list I can look into this more.

asaf-erlich (approved):
lgtm for now, there is still some commented code left in we should clean up eventually but that's fine.

PR #21, opened 2020-07-16
added modifier lists, changed tests to accomodate nested comparisons

jacobrobertbaca:
# Description of change
added stream modifier-lists
changed tests to accomodate nested comparisons

# Manual QA steps
 - checked manually that records came through
 - added to tests and ran tap-tester
 
# Risks
 - low
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -110,6 +110,11 @@ class TestSquareBase(unittest.TestCase):
                 self.REPLICATION_METHOD: self.INCREMENTAL,
                 self.REPLICATION_KEYS: {'updated_at'}
             },
+            "modifier_lists": {
+                self.PRIMARY_KEYS: {'id'},
+                self.REPLICATION_METHOD: self.INCREMENTAL,
+                self.REPLICATION_KEYS: {'updated_at'}
+            },
         }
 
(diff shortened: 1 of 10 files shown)


tests/base.py
```diff
@@ -225,16 +230,39 @@ def expected_schema_keys(self, stream):
...
+                        self.modify_expected_record(item)
+                    self.sort_array_type(expected_record, key, value)
+                else:
+                    self.align_date_type(expected_record, key, value)
 
-        for key, value in expected_records.items(): # Modify a single record
-            self.align_date_type(expected_records, key, value)
 
     def align_date_type(self, record, key, value):
         """datetime values must conform to ISO-8601 or they will be rejected by the gate"""
```
kspeer825, line 268: Ooops we still have to do this 1st To Do.

kspeer825: It can wait but I think there will be some datetime schema issues revealed once this is done.

kspeer825 (approved):
lgtm I put that one comment, your call if you want to address that together. I can do it on my test branch after this is merged if you prefer that.

PR #22, opened 2020-07-16
Testing/payments refunds

kspeer825:
# Description of change
Expand test client to creates and updates for payments and refunds.
Add payments and refunds to remaining tests.

# Manual QA steps
 - Verify create/update works as expected
 
# Risks
 - Low, Schema fix and test changes
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -65,16 +65,8 @@ class TestSquareBase(unittest.TestCase):
             'client_secret': os.getenv('TAP_SQUARE_APPLICATION_SECRET'),
         }
 
-    @staticmethod
-    def expected_check_streams():
-        return {
-            'items',
-            'categories',
-            'discounts',
-            'taxes',
-            'employees',
-            'locations',
-        }
+    def expected_check_streams(self):
+        return set(self.expected_metadata().keys()).difference(set())
 
     def expected_metadata(self):
         """The expected streams and metadata about the streams"""
@@ -107,6 +99,16 @@ class TestSquareBase(unittest.TestCase):
             "locations": {
                 self.PRIMARY_KEYS: {'id'},
                 self.REPLICATION_METHOD: self.FULL,
+            },
+            "refunds": {
+                self.PRIMARY_KEYS: {'id'},
+                self.REPLICATION_METHOD: self.INCREMENTAL,
+                self.REPLICATION_KEYS: {'created_at'}
+            },
+            "payments": {
+                self.PRIMARY_KEYS: {'id'},
+                self.REPLICATION_METHOD: self.INCREMENTAL,
+                self.REPLICATION_KEYS: {'updated_at'}
             },
         }
 
@@ -237,7 +239,7 @@ class TestSquareBase(unittest.TestCase):
     def align_date_type(self, record, key, value):
         """datetime values must conform to ISO-8601 or they will be rejected by the gate"""
         # TODO update this to execute fo all datetime objects
-        if isinstance(value, str) and key in ['updated_at']:
+        if isinstance(value, str) and key in ['updated_at', 'created_at']:# TODO: update to reflect replication keys
             raw_date = self.parse_date(value)
             iso_date = dt.strftime(raw_date,  "%Y-%m-%dT%H:%M:%S.%fZ")
             record[key] = iso_date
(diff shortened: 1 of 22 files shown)


jacobrobertbaca (approved):

tests/base.py
```diff
@@ -263,11 +263,10 @@ def modify_expected_record(self, expected_record):
                 else:
                     self.align_date_type(expected_record, key, value)
 
-
     def align_date_type(self, record, key, value):
         """datetime values must conform to ISO-8601 or they will be rejected by the gate"""
-        # TODO update this to execute fo all datetime objects
-        if isinstance(value, str) and key in ['updated_at', 'created_at']:# TODO: update to reflect replication keys
+        if isinstance(value, str) and isinstance(self.parse_date(value), dt):
```
asaf-erlich: Can we use type hints to avoid at least the first isinstance check?

tests/base.py
```diff
@@ -221,7 +221,7 @@ def parse_date(self, date_value):
                 date_stripped = dt.strptime(date_value, "%Y-%m-%dT%H:%M:%SZ")
                 return date_stripped
             except ValueError:
-                raise NotImplementedError
+                return date_value
```
asaf-erlich, line 255: Why would we return the date_value here and not just let the error bubble up? In fact, I'm a bit confused on this method for parse_date. It tries to parse in one format, if that fails parses in another? That seems like a bad code smell to me. Are we using it in different places that should be their own method for different fields?

tests/base.py
```diff
@@ -230,19 +230,22 @@ def expected_schema_keys(self, stream):
 
         return props.keys()
 
+    # TODO Determine if sorting is even a valid modifier for our expectations
```
asaf-erlich, line 264: @luandy64 and I spent some time trying to avoid sorting. We modified in my PR the code to use assertDictEqual, but it still relies on sorting. We investigated using a library that can compare dicts but it threw errors at one point unfortunately. I really wish python or the community had a better solution to this since it's a very common problem people have to deal with. See https://github.com/singer-io/tap-square/pull/25/files#diff-19f36877654ab85cc34581bf5817a1abR190 - We aught to refactor the tests because I suspect there's a lot of repetition here.

tests/test_bookmarks.py
```diff
@@ -226,47 +239,50 @@ def test_run(self):
...
+                            replicated_updated_record = [record for record in second_sync_data
+                                                         if updated_record.get('id') == record.get('id')]
+                            print("ACTUAL RECORD: {}".format(replicated_updated_record[0]))
+                            for key in schema_keys:
+                                val = replicated_updated_record[0].get(key)
+                                e_val = updated_record.get(key)
+                                if e_val != val:
+                                    print("\nDISCREPANCEY | KEY {}: ACTUAL: {} EXPECTED {}".format(key, val, e_val))
+                        self.assertIn(updated_record, second_sync_data,
+                                      msg="Data discrepancy for the updated record.")
```
asaf-erlich: Yeah, see the code I mentioned above again referring to our PR: https://github.com/singer-io/tap-square/pull/25/files#diff-19f36877654ab85cc34581bf5817a1abR190

tests/base.py
kspeer825: Well everything is coming through this. So adding type hints would cause the method to fail. We could add them and then do the type checking in the other method. But that would just move the `isinstance`

tests/base.py
kspeer825: I broke this method into 2. One that passes the record back if it wasn't a datetime. The other returns the error. The one that passes records through is used by the modify record method in base. It passes all types in and only parses the ones that fit this datetime formatting. If we are missing a format it will show up as a data discrepancy in the tests, and it can be added later.

tests/base.py
kspeer825: I used assertDictEqual in the other tests that are also comparing records line by line. This is very useful assertion.

kspeer825: I have used Collections while debugging which is good when you have a suspected duplicate record.

tests/test_bookmarks.py
kspeer825: Done.

PR #23, opened 2020-07-16
Bank accounts

jacobrobertbaca:
# Description of change
added bank accounts stream.
IMPORTANT: this endpoint only works in the prod app so we can't include it in any of the tap-tester tests (since that relies on writing to the sandbox)

# Manual QA steps
 - verified that record came through manually
 
# Risks
 - a little, we can't stress test it since we can't run tests against it. It's important for us to mention in the docs that it's not supported in some way
 
# Rollback steps
 - revert this branch


tap_square/client.py
```diff
@@ -96,6 +97,7 @@ def get_employees(self, bookmarked_cursor):
             yield (result.body.get('employees', []), result.body.get('cursor'))
 
     def get_locations(self):
+        body = {}
```
asaf-erlich, line 100: What is the purpose of storing this as a variable that you keep after the loop where it's used later?

tap_square/client.py
```diff
@@ -105,14 +107,36 @@ def get_locations(self):
         yield (result.body.get('locations', []), result.body.get('cursor'))
 
         while result.body.get('cursor'):
+            body['cursor'] = result.body.get('cursor')
```
asaf-erlich, line 110: I don't see why to store this when you can just inline the variable in the result line in line number 112 below.

tap_square/client.py
```diff
@@ -105,14 +107,36 @@ def get_locations(self):
...
-                result = self._client.locations.list_locations()
+                result = self._client.locations.list_locations(**body)
 
             if result.is_error():
                 raise Exception(result.errors)
 
             yield (result.body.get('locations', []), result.body.get('cursor'))
 
+    def get_bank_accounts(self):
+        body = {}
```
asaf-erlich, line 120: Again, why keep a body dict here?

asaf-erlich (approved):
Minor comments, but if tests are passing we should probably merge this and move on since you're OOO today

kspeer825:
These changes were made by the merging of branch `testing/bank_accounts`

PR #26, opened 2020-07-21
Fix locations schema

luandy64:
# Description of change
https://stitchdata.atlassian.net/browse/SRCE-3532

This PR adds fields to the `locations` schema. I edited a location in the Square Sandbox and used `singer-infer-schema`. `business_hours` was not showing up though, so  I referenced [this doc](https://developer.squareup.com/reference/square/locations-api/list-locations) to manually add the schema

# Manual QA steps
 - Ran the tests
 
# Risks
 - Low
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -296,6 +296,7 @@ class TestSquareBase(unittest.TestCase):
                     self.sort_array_type(expected_record, key, value)
                 else:
                     self.align_date_type(expected_record, key, value)
+                    self.align_number_type(expected_record, key, value)
 
     def align_date_type(self, record, key, value):
         """datetime values must conform to ISO-8601 or they will be rejected by the gate"""
@@ -304,6 +305,15 @@ class TestSquareBase(unittest.TestCase):
             raw_date = self.date_check_and_parse(value)
             iso_date = dt.strftime(raw_date,  "%Y-%m-%dT%H:%M:%S.%fZ")
             record[key] = iso_date
+
+    def align_number_type(self, record, key, value):
+        """float values must conform to json number formatting so we convert to Decimal"""
+        if isinstance(value, float) and key in ['latitude', 'longitude']:
+            try:
+                record[key] = str(value)
+            except Exception as ex:
+                print("Failed to convert {} from type float to string".format(value))
+                raise
 
     def sort_array_type(self, record, key, value):
         """
--- a/tests/base.py
+++ b/tests/base.py
@@ -347,11 +347,7 @@ class TestSquareBase(unittest.TestCase):
     def align_number_type(self, record, key, value):
         """float values must conform to json number formatting so we convert to Decimal"""
         if isinstance(value, float) and key in ['latitude', 'longitude']:
-            try:
-                record[key] = str(value)
-            except Exception as ex:
-                print("Failed to convert {} from type float to string".format(value))
-                raise
+            record[key] = str(value)
 
     def sort_array_type(self, record, key, value):
         """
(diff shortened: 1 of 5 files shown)


tests/base.py
```diff
@@ -343,6 +344,15 @@ def align_date_type(self, record, key, value):
             iso_date = dt.strftime(raw_date,  "%Y-%m-%dT%H:%M:%S.%fZ")
             record[key] = iso_date
 
+    def align_number_type(self, record, key, value):
+        """float values must conform to json number formatting so we convert to Decimal"""
+        if isinstance(value, float) and key in ['latitude', 'longitude']:
+            try:
+                record[key] = str(value)
+            except Exception as ex:
+                print("Failed to convert {} from type float to string".format(value))
```
asaf-erlich: What is the purpose of this print? Wouldn't the exception have all the info required?

tests/test_client.py
```diff
@@ -76,7 +76,7 @@ def get_all(self, stream, start_date=None, end_date=None):
             return [obj for page, _ in self.get_refunds(start_date, None) for obj in page]
         elif stream == 'payments':
             if not self.PAYMENTS:
-                return [obj for page, _ in self.get_payments(start_date, None) for obj in page]
+                self.PAYMENTS += [obj for page, _ in self.get_payments(start_date, None) for obj in page]
```
asaf-erlich: Since the above line is if not self.PAYMENTS it would be empty. So this should just be self.PAYMENTS =, not += right?

tests/base.py
kspeer825: Ah I don't think that is really needed now. It was there because for about 5 minutes because I was converting float to Decimal and wanted to log in case it failed. Don't think there will be an issue going from float to string.

tests/test_client.py
kspeer825: It can be either I guess, since it is instantiated to an empty string. Equals probably makes more sense.

kspeer825: * empty list

kspeer825: not string

asaf-erlich: I know it could be either, but I think it's confusing because the reader of the code has to stop and think for a second about what that might mean...

kspeer825: Changed.

tests/base.py
kspeer825: Removed.

tests/base.py
```diff
@@ -343,6 +344,11 @@ def align_date_type(self, record, key, value):
             iso_date = dt.strftime(raw_date,  "%Y-%m-%dT%H:%M:%S.%fZ")
             record[key] = iso_date
 
+    def align_number_type(self, record, key, value):
```
asaf-erlich, line 347: Not a big deal, but why call the parent method "modify_expected_record" but the specific record modifiers "align_<blank>_type"? shouldn't both be modify or align?

tests/base.py
```diff
@@ -334,6 +334,7 @@ def modify_expected_record(self, expected_record):
                     self.sort_array_type(expected_record, key, value)
                 else:
                     self.align_date_type(expected_record, key, value)
+                    self.align_number_type(expected_record, key, value)
```
asaf-erlich, line 337: I feel like this method is inconsistent. It goes if type is dict, do something, list do something, but then else and then each of the methods checks the type again. Why not put the type checking inside the modify part and then you don't have to do the type checking inside the methods?

kspeer825: This whole method needs to be refactored, the recursion was implemented with sorting records in mind which is not a valid modifier. We can address this in a future PR. I will make a ticket to clean up all tests and include this point so that this does not get forgotten.

asaf-erlich (approved):

PR #27, opened 2020-07-23
Add Orders

luandy64:
# Description of change
This PR adds Orders as a stream we can sync. It also updates the tests to account for different API limits

# Manual QA steps
 - Ran all tests locally and in Circle
 
# Risks
 - Low
 
# Rollback steps
 - revert this branch


kspeer825 (approved):
This looks good, if possible at the end we want to refactor Orders and remove the import of that stream object in the test client.

PR #29, opened 2020-07-23
Change refunds to be a full table stream since you can only query on created time but records can be updated

asaf-erlich:
# Description of change
Change refunds to be a full table stream since you can only query on created time but records can be updated, addresses:
https://stitchdata.atlassian.net/browse/SRCE-3591

# Manual QA steps
 - unit / integ tests
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_bookmarks.py
+++ b/tests/test_bookmarks.py
@@ -18,15 +18,20 @@ class TestSquareIncrementalReplication(TestSquareBase):
         return self.dynamic_data_streams().difference(
             {  # STREAMS NOT CURRENTY TESTABLE
                 'employees', # Requires production environment to create records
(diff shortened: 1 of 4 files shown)


tests/test_bookmarks.py
```diff
@@ -269,7 +274,7 @@ def test_run(self):
                 for primary_key in primary_keys:
 
                     # Verify that the inserted records are replicated by the 2nd sync and match our expectations
-                    schema_keys = set(self.expected_schema_keys(stream))
+                    set(self.expected_schema_keys(stream))
```
kspeer825: I am not sure what this is doing but I don't think we need it.

asaf-erlich: Ok I'll remove it.

luandy64 (approved):

PR #32, opened 2020-07-29
Testing/inventory

kspeer825:
# Description of change
Changes to inventories stream. 
Updated tests for use of creates and updates for inventories in test_client.

# Manual QA steps
 - 
 
# Risks
 - Medium, significant change to the `inventories` sync method.
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_square/streams.py
+++ b/tap_square/streams.py
@@ -155,7 +155,7 @@ class Orders():
 
 class Inventories:
     tap_stream_id = 'inventories'
-    key_properties = ['catalog_object_id', 'location_id']
+    key_properties = []
     replication_method = 'FULL_TABLE'
     valid_replication_keys = []
     replication_key = None
--- a/tap_square/streams.py
+++ b/tap_square/streams.py
@@ -161,11 +161,8 @@ class Inventories:
     replication_key = None
 
     def sync(self, client, start_time, bookmarked_cursor): #pylint: disable=no-self-use
-        items = Items()
-        all_variation_ids = set(items.get_all_variation_ids(client, start_time, bookmarked_cursor))
-
         # TODO: Need to pass in bookmarked_cursor into get_inventories
-        for page, cursor in client.get_inventories(all_variation_ids, start_time):
+        for page, cursor in client.get_inventories(start_time):
             yield page, cursor
 
 
--- a/tap_square/streams.py
+++ b/tap_square/streams.py
@@ -160,9 +160,9 @@ class Inventories:
     valid_replication_keys = []
     replication_key = None
 
-    def sync(self, client, start_time, bookmarked_cursor): #pylint: disable=no-self-use
+    def sync(self,client, start_time, bookmarked_cursor): #pylint: disable=no-self-use
         # TODO: Need to pass in bookmarked_cursor into get_inventories
-        for page, cursor in client.get_inventories(start_time):
+        for page, cursor in client.get_inventories(start_time, bookmarked_cursor):
             yield page, cursor
 
 
--- a/tap_square/streams.py
+++ b/tap_square/streams.py
@@ -161,7 +161,6 @@ class Inventories:
     replication_key = None
 
     def sync(self, client, start_time, bookmarked_cursor): #pylint: disable=no-self-use
-        # TODO: Need to pass in bookmarked_cursor into get_inventories
         for page, cursor in client.get_inventories(start_time, bookmarked_cursor):
             yield page, cursor
 
--- a/tap_square/streams.py
+++ b/tap_square/streams.py
@@ -160,7 +160,7 @@ class Inventories:
     valid_replication_keys = []
     replication_key = None
 
-    def sync(self,client, start_time, bookmarked_cursor): #pylint: disable=no-self-use
+    def sync(self, client, start_time, bookmarked_cursor): #pylint: disable=no-self-use
         # TODO: Need to pass in bookmarked_cursor into get_inventories
         for page, cursor in client.get_inventories(start_time, bookmarked_cursor):
             yield page, cursor
--- a/tap_square/streams.py
+++ b/tap_square/streams.py
@@ -155,7 +155,7 @@ class Orders():
 
 class Inventories:
     tap_stream_id = 'inventories'
-    key_properties = []
+    key_properties = ['catalog_object_id', 'location_id']
     replication_method = 'FULL_TABLE'
     valid_replication_keys = []
     replication_key = None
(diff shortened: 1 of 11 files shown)


tap_square/streams.py
```diff
@@ -161,11 +161,8 @@ class Inventories:
     replication_key = None
 
     def sync(self, client, start_time, bookmarked_cursor): #pylint: disable=no-self-use
-        items = Items()
-        all_variation_ids = set(items.get_all_variation_ids(client, start_time, bookmarked_cursor))
-
         # TODO: Need to pass in bookmarked_cursor into get_inventories
```
luandy64: Let's delete this TODO

tests/test_all_fields.py
```diff
@@ -174,35 +175,64 @@ def all_fields_test(self):
...
+                    # Verify that actual records were in our expectations
+                    for actual_record in actual_records:
+                        if actual_record not in expected_records.get(stream):
+                            print("DATA DISCREPANCY:\n\nACTUAL RECORD:\n{}\n".format(actual_record))
+                            for record in expected_records.get(stream):
+                                if record.get('catalog_object_id') == actual_record.get('catalog_object_id') and \
+                                   record.get('location_id') == actual_record.get('location_id'):
+                                    print("EXPECTED_RECORDS:")
+                                    print(str(record))
+                        self.assertIn(actual_record, expected_records.get(stream))
```
luandy64, line 221: Oh cool, I don't think I've seen `assertIn` before

tests/test_bookmarks.py
```diff
@@ -190,9 +191,13 @@ def test_run(self):
...
             first_rec_version = first_rec.get('version')
-            updated_record = self.client.update(stream, first_rec_id, first_rec_version)
+            if stream == 'inventories':
+                updated_record = self.client.update(stream, obj_id=None, version=None, obj=first_rec)
+            else:
+                updated_record = self.client.update(stream, first_rec_id, first_rec_version)
             assert len(updated_record) > 0, "Failed to update a {} record".format(stream)
-            assert len(updated_record) == 1, "Updated too many {} records".format(stream)
+            if stream != 'inventories':
+                assert len(updated_record) == 1, "Updated too many {} records".format(stream)
```
luandy64, line 200: Can't this be added to the `else` block above?

tests/test_client.py
```diff
@@ -179,6 +180,82 @@ def create(self, stream, ext_obj=None, start_date=None):
...
+        catalog_obj_id = item_variation.get('id')
+
+        # Get a random location
+        all_locations = self.get_all('locations')
+        loc_id =  all_locations[random.randint(0, len(all_locations) - 1)].get('id')
+        from_state = 'IN_STOCK'  # inventory_obj.get('state')
+
+        # Adjustment logic
+        made_id = self.make_id('inventory')
+        if from_state == 'IN_STOCK':
```
luandy64, line 211: Isn't this always true?

tests/test_client.py
```diff
@@ -507,6 +639,61 @@ def update_locations(self, obj_id):
...
 
+    def update_inventory_adjustment(self, catalog_obj):
+        catalog_obj_id = catalog_obj.get('catalog_object_id')
+        loc_id = catalog_obj.get('location_id')
+
+        from_state = 'IN_STOCK'  # inventory_obj.get('state')
+
+        # Adjustment logic
+        made_id = self.make_id('inventory')
+        if from_state == 'IN_STOCK':
```
luandy64, line 650: Same as before: isn't this always true?

tests/test_pagination.py
```diff
@@ -97,7 +98,7 @@ def pagination_test(self):
                     location_id = [location['id'] for location in self.client.get_all('locations')][0]
                     for i in range(num_to_post):
                         new_objects.append(self.client.create_order(location_id))
-                elif self.API_LIMIT.get(stream) < self.BATCH_LIMIT: # not all streams have batch endpoints
+                elif self.API_LIMIT.get(stream) < self.BATCH_LIMIT or stream == 'inventories': # not all streams have batch endpoints
```
luandy64: I see how `elif self.API_LIMIT.get(stream) < self.BATCH_LIMIT:` worked before. But I think we should change this to be more explicit. The `BATCH_LIMIT` is for the Catalog Object streams, but I'm okay with reusing it since `inventories` also has a limit of 1000. But we have 3 groups of streams now: Orders, Not Catalog Object Streams, and Catalog Object Streams

I think it should look something like this
```
if stream == 'orders':
  pass
elif stream in {'a', 'b', 'c'}: # The not-catalog-object streams
  pass
elif stream in {'d', 'e', 'f'}: # The catalog object ones
  pass
else:        # I could live with or without this
  raise Error
```

luandy64 (changes requested):
This is looking good. I have some things I'd like cleaned up

tap_square/streams.py
kspeer825: Done

tests/test_bookmarks.py
kspeer825: Done

tests/test_client.py
kspeer825: For the moment yes. Ideally we will change it to do different types of transitions.

tests/test_client.py
kspeer825: Same as previous comment

tests/test_pagination.py
kspeer825: Good call I thinks that is more clear.

luandy64:
Another review point I have, but maybe this is more stylistic, so I left it out of the review. I would like to hear thoughts about it if anyone has any.

```
primary_keys = self.expected_primary_keys().get(stream)
pk = list(primary_keys)[0] if primary_keys else None

if pk:
  pass
else:
  pass
```

This pattern shows up in the PR a few times. I would prefer this to check what stream we are dealing with. That way we know that `self.expected_primary_keys().get(stream)` will return something and we don't have to do this `if primary_keys else None`. 

This would also help with the readability of the test since we already do other `if stream == "blah"` in these tests

tests/test_pagination.py
kspeer825: Done

tests/test_bookmarks.py
luandy64: I think we can simplify this to:
```
if stream == 'inventories':
    updated_record = self.client.update(stream, obj_id=None, version=None, obj=first_rec)
else:
    updated_record = self.client.update(stream, first_rec_id, first_rec_version)
    assert len(updated_record) == 1, "Updated too many {} records".format(stream)
assert len(updated_record) > 0, "Failed to update a {} record".format(stream)
```

kspeer825: I like the ordering I have because if we fail to get a value returned for a standard stream, we will end up failing with `Updated too many <stream-name> streams` rather than the `Failed to update a <stream-name> record` message.

kspeer825: Which is not the case

luandy64 (approved):

PR #33, opened 2020-07-29
Add timecards

luandy64:
# Description of change
Adds `Shifts` as a stream

# Manual QA steps
 - Ran tests locally
 
# Risks
 - Low
 
# Rollback steps
 - revert this branch


tap_square/client.py
```diff
@@ -207,6 +207,33 @@ def get_inventories(self, variation_ids, start_time):
...
+                }
+            }
+        }
+        with singer.http_request_timer('GET shifts'):
+            result = self._client.labor.search_shifts(body=body)
+
+        if result.is_error():
+            raise Exception(result.errors)
+
+        yield (result.body.get('shifts', []), result.body.get('cursor'))
```
asaf-erlich, line 225: I don't like the pattern of doing the yield twice. It seems unnecessary. But we need to refactor a ton of this so it's not a high priority right this second.

tap_square/streams.py
```diff
@@ -169,6 +169,22 @@ def sync(self, client, start_time, bookmarked_cursor): #pylint: disable=no-self-
...
 
 
+class Shifts:
+    tap_stream_id = 'shifts'
+    key_properties = ['id']
+    replication_method = 'INCREMENTAL' # Maybe 'FULL_TABLE'
+    valid_replication_keys = ['updated_at']
+    # TODO We cannot query the API with an updated at
+    # - There's a `filter` in the `body` we can use, but that will filter on the start time of the shift
+    #   - Strategy 1: Always query with the start time for the tap config, ordering by `updated_at` and emit the
```
asaf-erlich, line 175: Should we ask the team for their thoughts on this? If you already wrote the code it's fine, but otherwise I would prefer to avoid the work if the tap is not used by so many people.

tap_square/sync.py
```diff
@@ -33,7 +33,31 @@ def sync(config, state, catalog):
             start_time = singer.get_bookmark(state, tap_stream_id, replication_key, config['start_date'])
             bookmarked_cursor = singer.get_bookmark(state, tap_stream_id, 'cursor')
 
-            if stream_obj.replication_method == 'INCREMENTAL':
+            if tap_stream_id == 'shifts':
```
asaf-erlich, line 36: Maybe we should make this some other flag on the stream instead of hardcoding to the stream id 'shifts'?

tap_square/sync.py
```diff
@@ -33,7 +33,31 @@ def sync(config, state, catalog):
...
+                state = singer.write_bookmark(
+                    state,
+                    tap_stream_id,
+                    replication_key,
+                    singer.utils.strftime(singer.utils.now(),
+                                          format_str=singer.utils.DATETIME_PARSE)
+                )
+                singer.write_state(state)
+
+            elif stream_obj.replication_method == 'INCREMENTAL':
```
asaf-erlich, line 60: Can we change the string of FULL_TABLE/INCREMENTAL to an enum?

tap_square/sync.py
```diff
@@ -33,7 +33,31 @@ def sync(config, state, catalog):
             start_time = singer.get_bookmark(state, tap_stream_id, replication_key, config['start_date'])
```
asaf-erlich, line 33: In general this sync method needs refactoring now that we're branching 3 times. There's definitely some repeated codes in those branches.

tests/base.py
```diff
@@ -153,7 +153,12 @@ def expected_metadata(self):
             "bank_accounts": {
                 self.PRIMARY_KEYS: {'id'},
                 self.REPLICATION_METHOD: self.FULL,
-            }
+            },
+            "shifts": {
+                self.PRIMARY_KEYS: {'id'},
+                self.REPLICATION_METHOD: self.INCREMENTAL,
+                self.REPLICATION_KEYS: {'updated_at'}
```
asaf-erlich, line 160: Does this work?

tests/test_bookmarks.py
```diff
@@ -177,7 +176,7 @@ def test_run(self):
             first_rec = first_sync_records.get(stream).get('messages')[0].get('data')
             first_rec_id = first_rec.get('id')
             first_rec_version = first_rec.get('version')
-            updated_record = self.client.update(stream, first_rec_id, first_rec_version)
+            updated_record = self.client.update(stream, first_rec_id, first_rec_version, obj=first_rec)
```
asaf-erlich: `obj` is too generic. Can we rename it to something else? Even `stream_data` seems better to me.

tests/test_client.py
```diff
@@ -24,6 +24,8 @@
 
 
 class TestClient(SquareClient):
+    SHIFT_MINUTES = 2
```
asaf-erlich, line 29: Maybe let's add a comment about why we chose 2 minutes (because shifts can't overlap).

tests/test_client.py
```diff
@@ -173,6 +177,25 @@ def create(self, stream, ext_obj=None, start_date=None):
             return [self.create_refunds(ext_obj, start_date).body.get('refund')]
         elif stream == 'payments':
             return [self.create_payments()]
+        elif stream == 'shifts':
+            employee_id = [employee['id'] for employee in self.get_all('employees')][0]
+            location_id = [location['id'] for location in self.get_all('locations')][0]
+
+            if not end_date:
```
asaf-erlich: I think we should make the parameter shift_minutes with it defaulting to self.SHIFT_MINUTES (might need to pull the constant out of the class to refer to it as a default value). It's unclear to the person looking at the method signature right now what the default behavior is.

tests/test_client.py
```diff
@@ -173,6 +177,25 @@ def create(self, stream, ext_obj=None, start_date=None):
...
+                end_date = singer.utils.strftime(end_date_datetime)
+
+            max_end_at = max([obj['end_at'] for obj in self.get_all('shifts')])
+            if start_date < max_end_at:
+                LOGGER.warning('Tried to create a Shift that overlapped another shift')
+                # Readjust start date and end date
+                start_date = max_end_at
+                start_date_parsed = singer.utils.strptime_with_tz(start_date)
+                end_date_datetime = start_date_parsed + timedelta(minutes = self.SHIFT_MINUTES)
+                end_date = singer.utils.strftime(end_date_datetime)
```
asaf-erlich, line 195: We copied pasted this code. Can we refactor to DRY it up?

tests/test_client.py
```diff
@@ -413,7 +455,7 @@ def create_batch_post(self, stream, num_records):
     ### UPDATEs
     ##########################################################################
 
-    def update(self, stream, obj_id, version):
+    def update(self, stream, obj_id, version, obj=None):
```
asaf-erlich: As I said earlier, can we change `obj` to something slightly less generic like `stream_data` or `stream_sample_record`?

tests/test_client.py
```diff
@@ -507,6 +551,26 @@ def update_locations(self, obj_id):
...
+                    "title": self.make_id('shift'),
+                    "hourly_rate" : obj['wage']['hourly_rate']
+                }
+            }
+        }
+        resp = self._client.labor.update_shift(id=obj['id'], body=body)
+
+        if resp.is_error():
+            raise RuntimeError(resp.errors)
+        LOGGER.info('Created a Shift with id %s', resp.body.get('shift',{}).get('id'))
```
asaf-erlich: Here the logger is after the resp.is_error() check, but in the create method it's before it. Can we be consistent (I slightly prefer it after).

tests/test_pagination.py
```diff
@@ -99,6 +100,20 @@ def pagination_test(self):
                     location_id = [location['id'] for location in self.client.get_all('locations')][0]
                     for i in range(num_to_post):
                         new_objects.append(self.client.create_order(location_id))
+                elif stream == 'shifts':
+                    # Find the max end_at to know when the last shift ends, so we can start a shift there
+                    max_end_at = max([obj['end_at'] for obj in existing_objects])
```
asaf-erlich, line 104: We're doing this logic twice, why?

tests/test_pagination.py
```diff
@@ -99,6 +100,20 @@ def pagination_test(self):
...
+                    end_at_datetime = utils.strptime_to_utc(max_end_at)
+                    for i in range(num_to_post):
+                        # Tested in the API Explorer that +00:00 works
+                        # How does this work? It feels like it should be new_objects += blah
```
asaf-erlich, line 108: What is this comment asking? `+=` concatenates two lists, `append` appends to the end of the list.

tests/test_pagination.py
```diff
@@ -99,6 +100,20 @@ def pagination_test(self):
                     location_id = [location['id'] for location in self.client.get_all('locations')][0]
                     for i in range(num_to_post):
                         new_objects.append(self.client.create_order(location_id))
+                elif stream == 'shifts':
+                    # Find the max end_at to know when the last shift ends, so we can start a shift there
+                    max_end_at = max([obj['end_at'] for obj in existing_objects])
+                    end_at_datetime = utils.strptime_to_utc(max_end_at)
```
asaf-erlich, line 105: Can we use type hints to avoid including the suffix `_datetime`?

tests/test_pagination.py
```diff
@@ -99,6 +100,20 @@ def pagination_test(self):
...
+                    for i in range(num_to_post):
+                        # Tested in the API Explorer that +00:00 works
+                        # How does this work? It feels like it should be new_objects += blah
+                        new_objects.append(
+                            # Create a shift that is self.client.SHIFT_MINUTES long
+                            self.client.create('shifts', start_date=utils.strftime(end_at_datetime))
+                        )
+                        # Bump our known max `end_at` by self.client.SHIFT_MINUTES
```
asaf-erlich, line 113: Mention that shifts cannot overlap in time?

tap_square/streams.py
luandy64: I'm partially wrote the code. I didn't do the part where we filter with the start time in the config, it would take a trivial amount of work to do so.

The tap currently orders by `updated_at` and will only write records with `updated_at` `>=` the bookmark

luandy64: We can ask the team and address the changes during the refactor

tap_square/client.py
luandy64: I agree that we can change this later

tests/test_client.py
luandy64: Done

tests/test_client.py
luandy64: We will address this later

tests/test_client.py
luandy64: I moved the log line in `create_shift()` and updated this one to not say `"Created a Shift... "`

tests/test_pagination.py
```diff
@@ -5,8 +5,8 @@
 import tap_tester.runner      as runner
 
 from datetime import timedelta, date
-from datetime import datetime as dt
-
+#from datetime import datetime as dt
```
asaf-erlich, line 8: Remove comment.

tests/test_pagination.py
asaf-erlich: Will remove comment in subsequent PR

asaf-erlich (approved):
LGTM

PR #34, opened 2020-07-30
Stream/v1 employee roles

jacobrobertbaca:
# Description of change
added roles stream
# Manual QA steps
 - added to test, ran tests
 
# Risks
 - low
 
# Rollback steps
 - revert this branch


tests/test_pagination.py
```diff
@@ -37,7 +37,8 @@ def testable_streams(self):
             {  # STREAMS NOT CURRENTY TESTABLE
                 'employees', # Requires production environment to create records
                 'modifier_lists',
-
+                'inventories',
```
asaf-erlich: Why was this added to not-testable?

tap_square/client.py
```diff
@@ -295,3 +302,46 @@ def get_payments(self, start_time, bookmarked_cursor):
...
             yield (result.body.get('payments', []), result.body.get('cursor'))
+
+
+    def get_batch_token(self, link): #pylint: disable=no-self-use
+        if link:
+            url = link[link.find('<')+1:link.find('>')]
+            parsed = urlparse.urlparse(url)
+            batch_token = parse_qs(parsed.query)['batch_token'][0]
+            return int(batch_token)
+        return None
```
asaf-erlich, line 313: I'm confused why this returns none and why the while loop works the way it does. What's the flow here?

asaf-erlich: Is it supposed to be the way it paginates?

tap_square/client.py
```diff
@@ -295,3 +302,46 @@ def get_payments(self, start_time, bookmarked_cursor):
...
+
+        yield (result.json(), batch_token)
+
+        while batch_token:
+            params['batch_token'] = batch_token
+            with singer.http_request_timer('GET payments'):
+                result = requests.get(url, headers=headers, params=params)
+
+            # if result.is_error():
+            #     raise Exception(result.errors)
```
asaf-erlich: Please uncomment this. All requests should check for errors and throw exception. Otherwise it hides failures.

tests/test_pagination.py
jacobrobertbaca: oh maybe it wasn't supposed to be there, it must have been a merge mistake.

tap_square/client.py
jacobrobertbaca: if there is no 'Link' in the header it means that the records are finished being returned, so i want batch_token to be None, because then it will break out of the while loop.

asaf-erlich: This actually makes sense now that i think about it. If we were using type hints this method would return an `Optional[int]` (Optional in python just means it's either an int or a None).

asaf-erlich, line 313: Can we change the order here to do if not link return None first? Like validate the input. Then do the url and batch token generation in the non-branch code?

asaf-erlich (approved):
lgtm with minor comment

tap_square/client.py
luandy64: Can we not use `find`? `requests` has utils functions for stuff like this

PR #38, opened 2020-07-30
Add using batch create for the inventories - seconds to create 1000 records instead of minutes

asaf-erlich:
# Description of change
Every 3 days it has to recreate 1000+ records, which took us minutes cause it repeats many calls (like get all locations) and also calls create or batch for 1 record * 1000, instead of batching it. This uses the batch apis correct.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -161,7 +161,7 @@ class TestClient(SquareClient):
 
     def create(self, stream, ext_obj=None, start_date=None, end_date=None):
         if stream == 'items':
-            return self.create_item().body.get('objects')
+            return self._create_item(start_date=start_date).body.get('objects')
         elif stream == 'categories':
             return self.create_categories().body.get('objects')
         elif stream == 'discounts':
@@ -173,7 +173,7 @@ class TestClient(SquareClient):
         elif stream == 'employees':
             return self.create_employees().body.get('objects')
         elif stream == 'inventories':
-            return self.create_batch_inventory_adjustment()
+            return self.create_batch_inventory_adjustment(start_date=start_date)
         elif stream == 'locations':
             return [self.create_locations().body.get('location')]
         elif stream == 'orders':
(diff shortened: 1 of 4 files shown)


tests/test_client.py
```diff
@@ -214,68 +215,75 @@ def get_an_inventory_adjustment(self, obj_id):
             raise RuntimeError('GET INVENTORY_ADJUSTMENT: {}'.format(response.errors))
         return response
 
-    def create_inventory_adjustment(self, start_date=None):
+    @staticmethod
+    def _chunks(lst, n):
```
luandy64: We should talk to QA, but I think we can import this from the tap

tests/test_client.py
```diff
@@ -214,68 +215,75 @@ def get_an_inventory_adjustment(self, obj_id):
...
+        """Yield successive n-sized chunks from lst."""
+        for i in range(0, len(lst), n):
+            yield lst[i:i + n]
+
+    def create_batch_inventory_adjustment(self, num_records):
         # Create an item
-        item = self.create_item().body.get('objects')[0]
+        #item = self.create_item().body.get('objects')[0]
```
luandy64: Let's delete this line at some point

tests/test_client.py
```diff
@@ -408,61 +393,43 @@ def create_modifier_list(self):
                                                                       'ordinal': 1}
                                                                  ],}
                                           }]}],
+
+    def create_item(self, num_records=1):
+        start_date = '2020-07-29T00:00:00Z'
```
luandy64: I think we said this, but I want to note it: we'll change this to use the start date in the test rather than be hard coded

luandy64 (approved):

tests/test_client.py
asaf-erlich: Did this, thanks

tests/test_client.py
asaf-erlich: Done

tests/test_client.py
asaf-erlich: ➕

asaf-erlich:
Pylint passes, tests are running, but fail in some asserts. Will fix those in a subsequent pr.

PR #41, opened 2020-07-31
Testing/mod lists pagination start date

kspeer825:
# Description of change
Add modifier_lists to pagination and start date.
Schema changes for modifier_lists and items streams.

# Manual QA steps
 - Tests ran locally.
 
# Risks
 - Low, only tap code change were for schema files.
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_square/client.py
+++ b/tap_square/client.py
@@ -1,12 +1,13 @@
 import urllib.parse as urlparse
 from urllib.parse import parse_qs
+import urllib.parse
 
 from datetime import timedelta
 from square.client import Client
 from singer import utils
 import singer
 import requests
-import urllib.parse
+
 
 LOGGER = singer.get_logger()
 
@@ -384,7 +385,7 @@ class SquareClient():
 
             yield (result.body.get('items', []), result.body.get('cursor'))
 
-    def get_settlements(self, location_id, start_time, bookmarked_cursor):
+    def get_settlements(self, location_id, start_time, bookmarked_cursor): #pylint: disable=unused-argument
 
         url = 'https://connect.squareup.com/v1/{}/settlements'.format(location_id)
         headers = {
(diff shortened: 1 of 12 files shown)


luandy64 (approved):

tap_square/client.py
```diff
@@ -311,6 +312,14 @@ def get_payments(self, start_time, bookmarked_cursor):
 
             yield (result.body.get('payments', []), result.body.get('cursor'))
 
+    def get_batch_token(self, link): #pylint: disable=no-self-use
+        if link:
+            url = link[link.find('<')+1:link.find('>')]
+            parsed = urllib.parse.urlparse(url)
+            batch_token = urllib.parse.parse_qs(parsed.query)['batch_token'][0]
+            return int(batch_token)
+        return None
```
asaf-erlich: Don't we already have this method used elsewhere?

tests/test_bookmarks.py
```diff
@@ -341,7 +343,12 @@ def test_run(self):
                 expected_records = expected_records_2.get(stream)
                 primary_keys = stream_primary_keys.get(stream)
                 pk = list(primary_keys)[0] if primary_keys else None
-                if stream != 'orders':  # ORDERS has too many dependencies to track explicitly
+                if stream in {'orders', 'modifier_lists'}:  # Some streams have too many dependencies to track explicitly
```
asaf-erlich: Later we should clean these lists of streams instead of maybe hardcoding them like this. Like maybe have constant sets at the top that describe what each one is / why. I bet if we look across different tests there are some patterns, but I could be wrong.

tap_square/client.py
kspeer825: Removed

tests/test_bookmarks.py
kspeer825: Agreed, I think this can be done in a later PR. Will add it to the refactor/cleanup ticket in Jira.

asaf-erlich (approved):

PR #42, opened 2020-08-04
Shift does query by start time filtering, but that is only by the cre…

asaf-erlich:
…ated_at of the shift, not the updated_at which it's bookmarked on

Also fixes the missing fields from the items schema, see the second jira below

# Description of change
- https://stitchdata.atlassian.net/browse/SRCE-3669
- https://stitchdata.atlassian.net/browse/SRCE-3606

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -161,10 +161,6 @@ class TestSquareBase(unittest.TestCase):
                 self.PRIMARY_KEYS: {'id'},
                 self.REPLICATION_METHOD: self.FULL
             },
-            "cash_drawer_shifts": {
-                self.PRIMARY_KEYS: {'id'},
-                self.REPLICATION_METHOD: self.FULL
-            },
             "roles": {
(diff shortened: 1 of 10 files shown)


tests/base.py
```diff
@@ -161,10 +161,6 @@ def expected_metadata(self):
                 self.PRIMARY_KEYS: {'id'},
                 self.REPLICATION_METHOD: self.FULL
             },
-            "cash_drawer_shifts": {
-                self.PRIMARY_KEYS: {'id'},
-                self.REPLICATION_METHOD: self.FULL
-            },
```
asaf-erlich, line 167: This is duplicate code that probably got merged in when there was a conflict, same lines exist higher up in the map

dmosorast (approved):
We may want to consider removing shifts from the exclusion list in the pagination test. But as is, this PR looks ready.

PR #44, opened 2020-08-05
Testing/production

kspeer825:
# Description of change
Add employees and roles to the test_client.
Incorporate production existing tests.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,4 +1,5 @@
-version: 2
+version: 2.1
+
 jobs:
   build:
     docker:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -38,10 +38,10 @@ jobs:
                      --password=$SANDBOX_PASSWORD \
                      --client-id=50 \
                      tests
-      - slack/status:
+     - slack/status:
           fail_only: true
           only_for_branches: testing/production
-          webhook: webhook
+          webhook: https://hooks.slack.com/services/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
 
 orbs:
   slack: circleci/slack@3.4.2
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,6 +37,15 @@ jobs:
                      --password=$SANDBOX_PASSWORD \
                      --client-id=50 \
                      tests
+                       steps:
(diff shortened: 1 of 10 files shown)


.circleci/config.yml
```diff
@@ -1,4 +1,5 @@
 version: 2
+
```
kspeer825, line 2: These lines were accidental, I will remove them.

tests/test_client.py
```diff
@@ -720,6 +770,52 @@ def update_categories(self, obj_id, version):
...
+        uid = self.make_id('employee')[1:]
+        data = {'first_name': uid,
+                'last_name': obj.get('last_name')}
+
+        resp = requests.put(url=full_url, headers=self.get_headers(), json=data)
+        if resp.status_code != 200:
+            raise Exception(resp.text)
+        return resp.json()
+
+    def update_roles_v1(self, obj):
```
asaf-erlich, line 793: This and the above method are nearly identical. Can you create a method they both call? I think it'll save us time in the future if we ever have to make changes to the way we call v1 endpoints.

kspeer825: I can, should I wait to until settlements is done though? Rule of 3

kspeer825: Nevermind I'll just do it

tests/test_bookmarks.py
```diff
@@ -200,28 +223,28 @@ def test_run(self):
...
+            first_rec = dict()
+            for message in first_sync_records.get('payments').get('messages'):
+                if message.get('data')['status'] == 'APPROVED':
+                    first_rec = message.get('data')
+                    break
+
+            if not first_rec:
+                raise RuntimeError("Unable to find any any payment with status APPROVED")
+            first_rec_id = first_rec.get('id')
+            first_rec_version = first_rec.get('version')
```
asaf-erlich, line 238: We do this in another test right? Can we put a method in the base class that they all use?

asaf-erlich (approved):
made 1 comment, your choice to fix it in this pr or next one

tests/test_bookmarks.py
kspeer825: We only do updates in this test, I don't think this happens anywhere else.

PR #46, opened 2020-08-05
Added back inventories to test_bookmarks and make fixes to the test to make it pass

asaf-erlich:
# Description of change
Added back inventories to test_bookmarks and make fixes to the test to make it pass which addresses:

https://stitchdata.atlassian.net/browse/SRCE-3611

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_bookmarks.py
+++ b/tests/test_bookmarks.py
@@ -219,8 +219,8 @@ class TestSquareIncrementalReplication(TestSquareBase):
             assert len(updated_record) > 0, "Failed to update a {} record".format('payments')
             assert len(updated_record) == 1, "Updated too many {} records".format('payments')
 
-            expected_records_second_sync['payments'] += updated_record
-            updated_records['payments'] += updated_record
+            expected_records_second_sync['payments'] += updated_record[0]
+            updated_records['payments'] += updated_record[0]
 
         # adjust expectations for full table streams to include the expected records from sync 1
         for stream in self.expected_full_table_streams():
--- a/tests/test_bookmarks.py
+++ b/tests/test_bookmarks.py
@@ -43,25 +43,6 @@ class TestSquareIncrementalReplication(TestSquareBase):
     def tearDownClass(cls):
         print("\n\nTEST TEARDOWN\n\n")
 
-    def poll_for_updated_record(self, rec_id, start_date):
-        from time import sleep
-        all_payments = self.client.get_all('payments', self.START_DATE)
-        temp_rec = [payment for payment in all_payments if payment['id'] == rec_id]
-
-        for i in range(10):
-            print("Polling {} iteration of payments for record: id={}".format(i, rec_id))
-            sleep(2)
-            all_payments = self.client.get_all('payments', self.START_DATE)
-            records = [payment for payment in all_payments if payment['id'] == rec_id]
-            assert len(records) == 1
-            record = records[0]
-            if len(temp_rec[0].keys()) < len(record.keys()):
-                print("Poll Successful after {} iterations.".format(i))
-                print(set(temp_rec[0].keys()).symmetric_difference(set(record.keys())))
-                break
-
-        return records
-
     def run_sync(self, conn_id):
         """
         Run a sync job and make sure it exited properly.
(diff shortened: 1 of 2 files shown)


tests/test_client.py
```diff
@@ -745,7 +747,7 @@ def update_payment(self, obj_id: str, action=None):
             raise RuntimeError("Require non-blank obj_id, found {}".format(obj_id))
 
         if not action:
-            action = random.choice([ 'complete', 'cancel' ])
+            action = random.choice([ 'complete'])
```
kspeer825: Did we determine that the status does indicate the presence of `processing_fee`? If that's the case I think it would be better if we could build that logic into `get_a_payment` rather than remove the path here.

kspeer825: I have a cleanup branch for the tests right now so this could be done there if our objective right now is to just get tests green.

dmosorast: This should probably be reverted back, since the handling is up there.

tests/test_client.py
```diff
@@ -863,16 +865,15 @@ def update_locations(self, obj_id):
...
         catalog_obj_id = catalog_obj.get('catalog_object_id')
         loc_id = catalog_obj.get('location_id')
 
         from_state = 'IN_STOCK'  # inventory_obj.get('state')
 
         # Adjustment logic
-        made_id = self.make_id('inventory')
         if from_state == 'IN_STOCK':
-            states = ['SOLD', 'WASTE'] # SOLD_ONLINE
+            states = ['SOLD'] # SOLD_ONLINE
```
dmosorast: Maybe `WASTE` should be here as a comment.

tests/test_client.py
```diff
@@ -333,6 +334,7 @@ def _create_batch_inventory_adjustment(self, start_date, num_records=1):
 
             all_counts += response.body.get('counts')
 
+        assert (len(all_counts) == 1 * num_records), "num_records={}, but len(all_counts)={}, all_counts={}, len(all_counts) should be num_records".format(2 * num_records, len(all_counts), all_counts)
```
dmosorast: The `1 *` can be removed here :)

tests/test_client.py
```diff
@@ -322,7 +323,7 @@ def _create_batch_inventory_adjustment(self, start_date, num_records=1):
         for change_chunk in chunks(changes, 100):
             body = {
                 'changes': change_chunk,
-                'ignore_unchanged_counts': random.choice([True, False]),
+                'ignore_unchanged_counts': False,
```
dmosorast, line 301: Do we still want this one? Not sure if it had any effect.

tests/test_bookmarks.py
```diff
@@ -408,16 +392,26 @@ def bookmarks_test(self):
...
+                        elif stream == 'inventories':
+                            self.assertInventoriesEqual(updated_record, sync_record)
                         else:
                             self.assertDictEqual(updated_record, sync_record)
 
-    def assertPaymentsEqual(self, created_record, sync_record):
-        self.assertEqual(frozenset(created_record.keys()), frozenset(sync_record.keys()))
-        created_record_copy = deepcopy(created_record)
+    def assertInventoriesEqual(self, expected_record, sync_record):
+        self.assertEqual(frozenset(expected_record.keys()), frozenset(sync_record.keys()))
```
kspeer825: If we could add a brief doc-string to these assertion methods explaining that square is inconsistent with ___ so we account for that by relaxing our assertions I think that would be good.

dmosorast (changes requested):
Just a few changes

tests/test_client.py
asaf-erlich: I put it back, the test should work with either one now.

tests/test_bookmarks.py
asaf-erlich: Sure.

tests/test_client.py
asaf-erlich: I don't know, but literally telling telling the api to inconsistently report records seems like a bad idea to me. I want to leave it like this... unless @kspeer825 feels strongly otherwise.

kspeer825: Agreed, False always seems better.

dmosorast (approved):
As long as the tests pass, this looks good to me :+1:

PR #48, opened 2020-08-07
Make created_at and updated_at keys that can be slightly off in the test

asaf-erlich:
# Description of change
https://stitchdata.atlassian.net/browse/SRCE-3673

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_bookmarks.py
+++ b/tests/test_bookmarks.py
@@ -224,8 +224,6 @@ class TestSquareIncrementalReplication(TestSquareBase):
 
             updated_pk_values = {tuple([record.get(pk) for pk in primary_keys]) for record in updated_records[stream]}
             for record in expected_records_first_sync.get(stream, []):
-                if isinstance(record, list):
-                    LOGGER.info("stream %s with list record about to be tupled, record=%s", stream, record)
                 record_pk_values = tuple([record.get(pk) for pk in primary_keys])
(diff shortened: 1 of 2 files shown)


tests/test_bookmarks.py
```diff
@@ -387,30 +379,27 @@ def bookmarks_test(self):
...
+        else:
+            self.assertDictEqual(expected_record, sync_record)
 
-        # Square api workflow updates these values so they're a few seconds different between the time the record is created and the tap syncs, but other fields are the same
-        self.assertGreaterEqual(sync_record_copy.pop('calculated_at'),
-                                expected_record_copy.pop('calculated_at'))
-        self.assertDictEqual(expected_record_copy, sync_record_copy)
-
-    def assertPaymentsEqual(self, expected_record, sync_record):
+    def assertDictEqualWithOffKeys(self, expected_record, sync_record, off_keys=set()):
```
kspeer825, line 390: Can we add a comment explaining this method is only valid for full table streams and that it is a workaround to square? I am just worried this pattern will get picked up on a different tap.

kspeer825: Oh  I guess we already say this below.

kspeer825 (approved):
Looks good

dmosorast (approved):

PR #49, opened 2020-08-07
Testing/cleanup

kspeer825:
# Description of change
Cleanup testable streams diffs. 
Add tests to catch duplicate pages of data in pagination test.
Fix start date test to ensure full table streams respect start date.
# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_pagination.py
+++ b/tests/test_pagination.py
@@ -194,7 +194,35 @@ class TestSquarePagination(TestSquareBase):
                         stream_actual_record = stream_actual_records[0]
                         self.assertEqual(expected_record.get(pk), stream_actual_record)
 
-                    # TODO account for streams without pks (inventories)
+                elif stream == 'inventories': # We can not compare by pks for this stream so we make a less strict assertion
+
+                    # Verify the expected number of records were replicated
+                    self.assertEqual(len(expected_records.get(stream)), len(actual_records))
+
+                    # Verify the pages contain unique data sets (verify we don't duplicate pages)
+                    page_size = self.API_LIMIT.get(stream)
+                    num_full_pages = len(actual_records) // page_size
+                    num_pages = num_full_pages if len(actual_records) % page_size == 0 else num_full_pages + 1
+
+                    pages_to_compare = []
+                    for page in range(num_pages):  # break the data into pages
+                        start = page*page_size
+                        end = start + page_size - 1
+                        pages_to_compare.append(actual_records[start:end])
+
+                    instances = 0
+                    for page_data in pages_to_compare: # count instances of a page of data
+                        for page in pages_to_compare:
+                            if page_data == page:
+                                instances += 1
+
+                        # Verify there is only 1 instance of a given page of data in the target
+                        self.assertEqual(1, instances, msg="There are duplicate pages of data")
+
+                        instances = 0
+
+                else:
+                    raise NotImplementedError("{} does no have full test coverage for pagination feature.".format(stream))
 
 
 if __name__ == '__main__':
--- a/tests/test_pagination.py
+++ b/tests/test_pagination.py
@@ -39,8 +39,9 @@ class TestSquarePagination(TestSquareBase):
         return self.dynamic_data_streams().difference(self.untestable_streams())
 
     def testable_streams_static(self):
-        testable_streams = self.static_data_streams().difference(self.untestable_streams())
-        return testable_streams.difference({'locations'})  # Locations does not paginate
+        return self.static_data_streams().difference(self.untestable_streams()).difference({
+            'locations',  # This stream does not paginate in the sync (See Above)
+        })
 
     def test_run(self):
         """Instantiate start date according to the desired data set and run the test"""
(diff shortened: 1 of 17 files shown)


tests/test_pagination.py
```diff
@@ -152,16 +149,66 @@ def pagination_test(self):
...
-                                     set(), msg="A paginated synced stream has a record that is missing expected fields.")
+                    self.assertEqual(set(), auto_fields.difference(actual_keys),
+                                     msg="A paginated synced stream has a record that is missing expected fields.")
 
-                # TODO ADD CHECK ON IDS
                 # Verify by pks that the data replicated matches what we expect
+                primary_keys = self.expected_primary_keys().get(stream)
+                pk = list(primary_keys)[0] if primary_keys else None
+
+                if pk:
```
asaf-erlich: If this branching is really necessary can you pull that code inside into methods. There's a lot going on and I believe it could be easier to read that way.

tests/test_pagination.py
```diff
@@ -152,16 +149,66 @@ def pagination_test(self):
...
+                                        msg="An expected record is missing from the sync: \nRECORD: {}".format(expected_record))
+                        self.assertEqual(1, len(stream_actual_records),
+                                         msg="A duplicate record was found in the sync for {}.".format(stream))
+                        stream_actual_record = stream_actual_records[0]
+                        self.assertEqual(expected_record.get(pk), stream_actual_record)
+
+                elif stream == 'inventories': # We can not compare by pks for this stream so we make a less strict assertion
+
+                    # Verify the expected number of records were replicated
+                    self.assertEqual(len(expected_records.get(stream)), len(actual_records))
```
asaf-erlich: Couldn't we assert this in all cases and pull it out of the if block?

tests/test_pagination.py
```diff
@@ -152,16 +149,66 @@ def pagination_test(self):
...
+                        pages_to_compare.append(actual_records[start:end])
+
+                    instances = 0
+                    for page_data in pages_to_compare: # count instances of a page of data
+                        for page in pages_to_compare:
+                            if page_data == page:
+                                instances += 1
+
+                        # Verify there is only 1 instance of a given page of data in the target
+                        self.assertEqual(1, instances, msg="There are duplicate pages of data")
```
asaf-erlich: I do not understand what this is asserting. Why are we doing all this work to break things into pages?

kspeer825: The test case is `Verify the pages contain unique data sets (verify we don't duplicate pages)`. Dan or someone had mentioned at some point that there was a tap that was paginating but was returning the same page multiple times.

tests/test_pagination.py
kspeer825: Yes we could, but the assertions above will catch if the lengths are different and have the benefit of printing the extra/missing record. If we added this prior to the assertions above it would prevent us from logging the diff. If we added it after it would never fail because we would fail on line 180 or 169 first.

kspeer825: Well I suppose it doesn't hurt the test to pull it from the elif and change that to an if. I can do that

tests/test_pagination.py
```diff
@@ -71,6 +62,11 @@ def test_run(self):
         self.TESTABLE_STREAMS = self.testable_streams().difference(self.sandbox_streams())
         self.pagination_test()
 
+        print("\n\n-- SKIPPING -- TESTING WITH STATIC DATA IN SQUARE_ENVIRONMENT: {}".format(os.getenv('TAP_SQUARE_ENVIRONMENT')))
```
asaf-erlich, line 66: Shouldn't this no longer say skipping?

tests/test_pagination.py
```diff
@@ -152,16 +150,42 @@ def pagination_test(self):
...
+                    pks = self.makeshift_primary_keys().get(stream)
 
+                # Verify by pks that the replicated records match our expectations
+                self.assertRecordsEqualByPK(stream, expected_records.get(stream), actual_records, pks)
+
+                # Verify the expected number of records were replicated
+                self.assertEqual(len(expected_records.get(stream)), len(actual_records))
+
+
+    def assertRecordsEqualByPK(self, stream, expected_records, actual_records, pks):
```
asaf-erlich, line 169: Feels like we could reuse this in multiple tests and pull this into the base method

tests/test_pagination.py
```diff
@@ -152,16 +150,42 @@ def pagination_test(self):
...
+        expected_pks_set = set(expected_pks)
+        self.assertEqual(len(expected_pks), len(expected_pks_set), msg="Our expectations contain a duplicate record.")
+
+        # Verify that all expected records were replicated
+        self.assertEqual(set(), expected_pks_set.difference(actual_pks_set),
+                         msg="Our expectations have more unique records than the tap replicated.")
+
+        # Verify ONLY expected records were replicated
+        self.assertEqual(set(), actual_pks_set.difference(expected_pks_set),
+                         msg="The tap replicated data that was not in our expectations.")
```
asaf-erlich: Can you compare both sets are equal instead of comparing the difference is 0 both depending on whether you put one first or the other?

tests/test_start_date.py
```diff
@@ -202,40 +192,41 @@ def start_date_test(self):
...
+                if stream == 'inventories':
+                    continue
 
 
-                # Testing how INCREMENTAL streams handle start date
-                elif replication_type == self.INCREMENTAL:
+                # BUG | https://stitchdata.atlassian.net/browse/SRCE-3681
+                # Skipping these two streams until BUG resolved.
+                # NOTE: we skip inventories ^ for a different reason leave that as is
+                if stream in {'roles', 'employees', 'locations'}: # TODO REMOVE
```
asaf-erlich, line 212: Can you comment on the bug we need to remove this? I'm sure I'll forget otherwise

asaf-erlich (approved):
lgtm with minor comments

tests/test_pagination.py
kspeer825: The only static stream is `locations` now and that doesn't paginate. So I will remove the references and put a comment explaining that.

tests/test_pagination.py
kspeer825: I liked having the separate error messages, but I guess it is cleaner to have one assertion.

PR #51, opened 2020-08-10
Remove prod app from tests

dmosorast:
# Description of change
The square prod credentials are meant to be used for actual production use cases. The "production" Square account depends on the authorization URL used in the OAuth flow, and will be encoded in the refresh token.

# Manual QA steps
 - N/A, automation should catch this change.
 
# Risks
 - Low, it's tests only
 
# Rollback steps
 - revert this branch


asaf-erlich (approved):

dmosorast:
Closing this PR, since during testing it seems like the oauth ID and secret actually have an effect.

PR #52, opened 2020-08-11
Tests don't always create records unnecessarily

asaf-erlich:
# Description of change
Fix minor bugs in the test, refactor code, and pylint some test files (not all yet)

# Manual QA steps
 - Ran some tests manually
 
# Risks
 - 
 
# Rollback steps
 - revert this branch


tests/#test_start_date.py#
```diff
@@ -0,0 +1,244 @@
+import os
```
asaf-erlich: This file shouldn't be here, removing

tests/test_client.py
```diff
@@ -819,46 +853,15 @@ def update_roles_v1(self, obj):
...
-        resp = self._client.catalog.update_item_modifier_lists(body)
-        if resp.is_error():
-            raise RuntimeError("Unable to UPDATE modifier_lists: {}".format(resp.errors))
-        # Updated modifier_lists return only the 'updated_at' in the response
-        # so we are grabbing the response body from a get to copare
-        mod_lists = self.get_all('modifier_lists', start_date)
-        updated_mod_lists = [ml for ml in mod_lists if ml.get('id') == modifier_list_id]
-        assert len(updated_mod_lists) == 1
-        return updated_mod_lists
-
```
kspeer825, line 858: This shouldn't have been delete but I have it on my current branch so I think it's fine.

tests/test_client.py
```diff
@@ -232,7 +232,13 @@ def create(self, stream, start_date, end_date=None, num_records=1):
         elif stream == 'shifts':
             employee_id = [employee['id'] for employee in self.get_all('employees', start_date)][0]
             location_id = [location['id'] for location in self.get_all('locations', start_date)][0]
-            max_end_at = max([obj['end_at'] for obj in self.get_all('shifts', start_date)])
+            all_shifts = self.get_all('shifts', start_date)
+
+            if all_shifts:
+                max_end_at = max([obj['end_at'] for obj in all_shifts])
+            else:
+                max_end_at = start_date
```
kspeer825, line 240: I assume this means that shifts does not obey the start_date? Otherwise we won't be getting these reccords.

kspeer825 (approved):
these changes seem fine to me

.circleci/config.yml
```diff
@@ -15,10 +15,15 @@ jobs:
             pip install --upgrade pip
             pip install .[dev]
       - run:
-          name: 'pylint'
+          name: 'pylint tap'
```
dmosorast, line 18: Shouldn't this be `tap_square`?

tests/test_client.py
asaf-erlich: This method is not used anywhere so I removed it.

.circleci/config.yml
```diff
@@ -15,10 +15,15 @@ jobs:
...
+            pylint tap_square -d 'broad-except,chained-comparison,empty-docstring,fixme,invalid-name,line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,no-else-raise,no-else-return,too-few-public-methods,too-many-arguments,too-many-branches,too-many-lines,too-many-locals,ungrouped-imports,wrong-spelling-in-comment,wrong-spelling-in-docstring,too-many-public-methods'
+      #- run:
+      #    name: 'pylint tests'
+      #    command: |
+      #      source /usr/local/share/virtualenvs/tap-square/bin/activate
+      #      pylint tests -d 'broad-except,chained-comparison,empty-docstring,fixme,invalid-name,line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,no-else-raise,no-else-return,too-few-public-methods,too-many-arguments,too-many-branches,too-many-lines,too-many-locals,ungrouped-imports,wrong-spelling-in-comment,wrong-spelling-in-docstring,duplicate-code,no-name-in-module'
```
dmosorast, line 26: Do we want to pylint the tests? I feel like this should either be deleted or uncommented.

tests/test_client.py
asaf-erlich: I don't think that's true. The shifts do obey the start date. But if we are out of records the max method will fail for an empty list.

tests/base.py
```diff
@@ -236,13 +233,18 @@ def expected_replication_keys(self):
...
+                if properties.get(self.START_DATE_KEY)}
+
     def expected_automatic_fields(self):
         auto_fields = {}
         for k, v in self.expected_metadata().items():
             auto_fields[k] = v.get(self.PRIMARY_KEYS, set()) | v.get(self.REPLICATION_KEYS, set())
         return auto_fields
 
-    def select_all_streams_and_fields(self, conn_id, catalogs, select_all_fields: bool = True, exclude_streams=[]):
+    def select_all_streams_and_fields(self, conn_id, catalogs, select_all_fields: bool = True, exclude_streams=None):
```
dmosorast, line 247: Unless this is always passed in, it'll need a fallback `exclude_streams = exclude_streams or []` or something similar

tests/base.py
```diff
@@ -59,34 +60,31 @@ def set_environment(self, env):
...
+    def get_properties(self, original=True):
         # Default values
         return_value = {
-            'start_date' : dt.strftime(dt.utcnow()-timedelta(days=3), self.START_DATE_FORMAT),
-            'sandbox' : 'true' if self.get_environment() == self.SANDBOX else 'false'
+            'start_date': dt.strftime(dt.utcnow() - timedelta(days=3), self.START_DATE_FORMAT),
+            'sandbox': 'true' if self.get_environment() == self.SANDBOX else 'false'
         }
-        if self.get_environment() == self.PRODUCTION:
-            self.SQUARE_ENVIRONMENT = self.PRODUCTION
```
dmosorast, line 72: Did this just not get used?

tests/base.py
```diff
@@ -304,7 +309,8 @@ def parse_date(self, date_value):
             except ValueError:
                 raise NotImplementedError("We are not accounting for dates of this format: {}".format(date_value))
 
-    def date_check_and_parse(self, date_value):
+    @staticmethod
```
dmosorast, line 312: I feel like this is a code smell, but not sure what's rotting 😅

dmosorast (changes requested):
Some comments, maybe some changes?

dmosorast (approved):

.circleci/config.yml
asaf-erlich: I do intend to turn it on when I get all the pylint warnings squared away, but we can remove it if you feel strongly

tests/test_client.py
kspeer825: But if we default to a max_end_at equal to the start_date, won't all records exist prior to the start_date?

PR #53, opened 2020-08-12
Fixing start_date handling and added canary test to try to sync all streams

dmosorast:
# Description of change
https://stitchdata.atlassian.net/browse/SRCE-3681
Fixing start_date handling, aligning the patterns a bit better across tests, and added canary test to try to sync all streams.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch


tests/base.py
```diff
@@ -426,3 +456,74 @@ def create_test_data(self, testable_streams, start_date, start_date_2=None, min_
...
+        catalogs = menagerie.get_catalogs(conn_id)
+
+        # Ensure our selection worked
+        for cat in catalogs:
+            catalog_entry = menagerie.get_annotated_schema(conn_id, cat['stream_id'])
+            # Verify all testable streams are selected
+            selected = catalog_entry.get('annotated-schema').get('selected')
+            print("Validating selection on {}: {}".format(cat['stream_name'], selected))
+            if cat['stream_name'] not in self.testable_streams(environment, data_type):
+                self.assertFalse(selected, msg="Stream selected, but not testable.")
```
kspeer825, line 504: I don't love that we now have all these assertions in base. Typically we don't want to abstract away what we are testing from the test itself.

tests/base.py
```diff
@@ -223,6 +230,30 @@ def expected_full_table_streams(self):
...
+            if data_type == DataType.DYNAMIC:
+                return self.testable_streams_dynamic().intersection(self.sandbox_streams())
+            else:
+                return self.testable_streams_static().intersection(self.sandbox_streams())
+        else:
+            if data_type == DataType.DYNAMIC:
+                return self.testable_streams_dynamic().intersection(self.production_streams())
+            else:
+                return self.testable_streams_static().intersection(self.production_streams())
+
```
kspeer825, line 256: I think this is a much better solution to dealing with multiple environments and data sets.

tests/test_pagination.py
```diff
@@ -5,12 +5,13 @@
...
 
 import singer
+from unittest import TestCase
 from base import TestSquareBase
 
 LOGGER = singer.get_logger()
 
 
-class TestSquarePagination(TestSquareBase):
+class TestSquarePagination(TestSquareBase, TestCase):
```
kspeer825, line 14: Why is TestCase needed?

tests/test_client.py
```diff
@@ -700,15 +699,16 @@ def _create_orders(self, num_records, start_date):
...
+                },
+                'state': 'PROPOSED'
             }]
             created_orders.append(self.post_order(body, location_id).body.get('order'))
 
         return created_orders
 
     def create_shift(self, start_date, end_date, num_records):
-        employee_id = [employee['id'] for employee in self.get_all('employees', start_date)][0]
+        employee_id = self.get_first_found('employees', None)['id']
```
kspeer825, line 710: I think this abstracts away the create. I think a better pattern is to have the logic in the method, but I know that isn't very DRY. Maybe the method name could be changed so that we know it creates a record if it can't find one.

tests/test_client.py
```diff
@@ -355,21 +365,10 @@ def create_refund(self, start_date):
 
         refund = self._client.refunds.refund_payment(body)
         if refund.is_error():
-            print("Refund error, Updating payment status and retrying refund process.")
-
-            if "PENDING_CAPTURE" in refund.body.get('errors')[0].get('detail'):
-                payment = self._update_payment(obj_id=payment_id, action='complete') # update (complete) a payment if it is pending
```
kspeer825, line 361: Why did we drop this part?

tests/test_all_fields.py
```diff
@@ -65,65 +62,9 @@ def all_fields_test(self):
...
-        # Verify tap exit codes
-        exit_status = menagerie.get_exit_status(conn_id, sync_job_name)
-        menagerie.verify_sync_exit_status(self, exit_status, sync_job_name)
-
-        # read target output
-        first_record_count_by_stream = runner.examine_target_output_file(self, conn_id,
-                                                                         self.expected_streams(),
-                                                                         self.expected_primary_keys())
-        replicated_row_count =  sum(first_record_count_by_stream.values())
+        (_, first_record_count_by_stream) = self.run_initial_sync(environment, data_type)
```
kspeer825, line 65: I'm not sure about pulling all that code from the test. I think it should go back in.

jacobrobertbaca (approved):

tests/test_sync_canary.py
```diff
@@ -0,0 +1,37 @@
...
+
+    def test_run(self):
+        """
+        Verify that for each stream you can call sync just so it exercises the code.
+        """
+        print("\n\nRUNNING {}".format(self.name()))
+        print("WITH STREAMS: {}\n\n".format(self.expected_streams()))
+
+        self.set_environment(self.SANDBOX)
+        self.run_initial_sync(self.get_environment(), DataType.DYNAMIC)
```
kspeer825, line 32: Ah now I see why you made that method. I think if you want to do that here that is fine but it should be a method defined in the test, not in base.

tests/test_client.py
asaf-erlich: I'll rename it get_or_create_first_found in a subsequent PR

PR #54, opened 2020-08-17
Added backoff library around client methods + Refactoring

asaf-erlich:
# Description of change
Added backoff library around client methods + Refactoring
* https://stitchdata.atlassian.net/browse/SRCE-3680

# Manual QA steps
 - Manually ran sync test with throwing the exception in a loop to confirm with logged messages that it retried
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,20 +64,6 @@ jobs:
                      tests/test_sync_canary.py
       - run:
           when: always
-          name: 'Testing Start Date'
-          command: |
-            source dev_env.sh
-            source /usr/local/share/virtualenvs/tap-tester/bin/activate
-            PYTHONPATH=$PYTHONPATH:/usr/local/share/virtualenvs/tap-square/lib/python3.5/site-packages/ \
-            run-test --tap=tap-square \
-                     --target=target-stitch \
-                     --orchestrator=stitch-orchestrator \
-                     --email=harrison+sandboxtest@stitchdata.com \
-                     --password=$SANDBOX_PASSWORD \
-                     --client-id=50 \
-                     tests/test_start_date.py
-      - run:
-          when: always
           name: 'Testing Automatic Fields'
           command: |
             source dev_env.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,7 +18,7 @@ jobs:
           name: 'pylint tap'
           command: |
             source /usr/local/share/virtualenvs/tap-square/bin/activate
-            pylint tap_square -d 'broad-except,chained-comparison,empty-docstring,fixme,invalid-name,line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,no-else-raise,no-else-return,too-few-public-methods,too-many-arguments,too-many-branches,too-many-lines,too-many-locals,ungrouped-imports,wrong-spelling-in-comment,wrong-spelling-in-docstring,too-many-public-methods'
+            pylint tap_square -d '' # Removing all ignores works, so leaving it, put exceptions in here if necessary
       #- run:
       #    name: 'pylint tests'
       #    command: |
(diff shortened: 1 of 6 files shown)


asaf-erlich:
Woot, I even got logs of where it retried:

```
2020-08-17 16:14:15,487Z    tap - INFO Backing off _retryable_v1_method(...) for 0.5s (requests.exceptions.HTTPError: 503 Server Error: Service Unavailable for url: https://connect.squareup.com/v1/me/roles?batch_token=300)
        2020-08-17 16:14:15,488Z    tap - WARNING Network error receiving data from square. Sleeping 0.5 seconds before trying again
```

And it succeeded after that

.circleci/config.yml
```diff
@@ -62,20 +62,6 @@ jobs:
                      --password=$SANDBOX_PASSWORD \
                      --client-id=50 \
                      tests/test_sync_canary.py
-      - run:
-          when: always
-          name: 'Testing Start Date'
```
asaf-erlich, line 67: Note this is a duplicate that must have been merged in by 2 people, the test still runs in circle

tap_square/client.py
```diff
@@ -52,18 +64,33 @@ def _get_access_token(self):
...
+        max_time=180, # seconds
+        on_backoff=log_backoff,
+        jitter=backoff.full_jitter,
+    )
+    def _retryable_v2_method(request_method, body, **kwargs):
+        result = request_method(body, **kwargs)
+
+        if result.is_error():
+            error_message = result.errors if result.errors else result.body
+            if 'Service Unavailable' in error_message:
```
asaf-erlich, line 79: For v2 this is the text in the reason that we included in the JIRA for when we should retry, we don't want to retry on any error but we can add more cases here if we want in the future

tests/test_client.py
```diff
@@ -235,7 +235,7 @@ def create(self, stream, start_date, end_date=None, num_records=1): # pylint: di
         else:
             raise NotImplementedError("create not implemented for stream {}".format(stream))
 
-    def get_first_found(self, stream, start_date):
+    def get_or_create_first_found(self, stream, start_date):
```
asaf-erlich, line 239: This rename was from a previous PR feedback requested by @kspeer825  so I carried it over here

setup.py
```diff
@@ -11,7 +11,8 @@
       py_modules=['tap_square'],
       install_requires=[
           'singer-python==5.9.0',
-          'squareup==5.3.0.20200528'
+          'squareup==5.3.0.20200528',
+          'backoff==1.8.0',
```
luandy64, line 15: PyPI lists the most recent version as `1.10.0`, is there any reason we picked `1.8.0`?

tap_square/client.py
```diff
@@ -220,93 +247,89 @@ def get_payments(self, start_time, bookmarked_cursor):
...
-                limit=1000,
-            )
-
-        if result.is_error():
-            raise RuntimeError(result.errors)
-
-        yield (result.body.get('items', []), result.body.get('cursor'))
+        while cursor:
+            if cursor == '__initial__':
+                cursor = bookmarked_cursor
```
luandy64: Do we need this if block?

If we have a `bookmarked_cursor` passed in, then this gets skipped. If we do set this to `__initial__` above, then we hit this block and set `cursor = None` (or some other falsey value). But then we do nothing with `cursor` until we maybe overwrite it with `result.body.get('cursor')`, which could set it to `None` anyway.

luandy64 (commented):
Just a few questions and then I think this looks good

setup.py
asaf-erlich: singer-python lists 1.8.0 - pip install reports a conflict if I try to use the latest. The alternative is a PR to update singer-python library and release a new one, which I'm happy to do since it would be a minor version bump at the the most and change nothing really about the library

tap_square/client.py
asaf-erlich: In the latest version I added a comment "# initial value so while loop is always entered one time". It's set to something to force the while loop to go through at least once.

luandy64 (approved):
We talked about this, and everything looks good

tap_square/client.py
asaf-erlich: Fixed it, removed that block

PR #55, opened 2020-08-17
Missing fields in streams orders shifts + Fix start date test for all streams

asaf-erlich:
# Description of change
* https://stitchdata.atlassian.net/browse/SRCE-3681
* https://stitchdata.atlassian.net/browse/SRCE-3700
* https://stitchdata.atlassian.net/browse/SRCE-3704

# Manual QA steps
 - ran tests locally where necessary, but all tests run in circle anyway
 
# Risks
 - 
 
# Rollback steps
 - revert this branch


tests/base.py
```diff
@@ -160,7 +161,8 @@ def expected_metadata(self):
             },
             "roles": {
                 self.PRIMARY_KEYS: {'id'},
-                self.REPLICATION_METHOD: self.FULL
+                self.REPLICATION_METHOD: self.FULL,
+                self.START_DATE_KEY: 'updated_at'
```
kspeer825, line 165: Why can't we use created_at for roles or employees?

tests/test_start_date.py
```diff
@@ -201,33 +196,27 @@ def start_date_test(self):
...
-                for record in records_from_sync_2:
-                    self.assertGreaterEqual(self.parse_date(record), self.parse_date(self.START_DATE_2),
-                                            msg="Record was created prior to start date for 2nd sync.\n" +
-                                            "Sync 2 start_date: {}\n".format(self.START_DATE_2) +
-                                            "Record bookmark: {} ".format(record))
+                self.assertRecordsDateGreaterEqual(synced_records_1, stream, self.START_DATE_1)
+                self.assertRecordsDateGreaterEqual(synced_records_2, stream, self.START_DATE_2)
+
+    def assertRecordsDateGreaterEqual(self, sync_records, stream, start_date):
+        replication_type = self.expected_replication_method().get(stream)
```
kspeer825: The replication_type is already defined on 187. I think it would actually make sense to move the start_date_key logic from below out of this assertion up to where replication type is defined at the start of the loop. And then just add it as a parameter to this method. Just a suggestion, not a huge deal either way.

tests/base.py
asaf-erlich: updated_at is always preferred, why would we want it to be created_at? What if a record older than the created_at date is updated?

tests/test_start_date.py
asaf-erlich: I pulled it into a base method since the logic for determine start date is also used elsewhere in the base. But I don't entirely agree. If it's not used anywhere else why pull it out of the method?

kspeer825: Well this pattern of abstracting assertions into methods isn't one we typically use. I think it's fine as long as it is clear. I think it is more clear now.

tests/base.py
kspeer825: That makes sense

tests/base.py
```diff
@@ -527,3 +554,68 @@ def run_initial_sync(self, environment, data_type):
...
+
+            if 'fulfillments' in sync_record_copy:
+                expected_record_fulfillment = expected_record_copy['fulfillments'][0]
+                synced_record_fulfillment = sync_record_copy['fulfillments'][0]
+                self.assertParentKeysEqual(expected_record_fulfillment, synced_record_fulfillment)
+
+                if 'pickup_details' in synced_record_fulfillment:
+                    expected_record_pickup_details = expected_record_fulfillment['pickup_details']
+                    if 'expired_at' in expected_record_pickup_details:
+                        # No explanation for why this key exists in the expected_record_pickup_details but not in synced record's
```
asaf-erlich: After thinking about this it's probably a missing field in the schema... great

asaf-erlich: Confirmed the schema for pickup_details has no expired_at field... this one took a while to debug.

tap_square/streams.py
```diff
@@ -1,15 +1,20 @@
 import singer
 
+LOGGER = singer.get_logger()
```
luandy64, line 3: It looks like we don't use this, so should we remove it?

luandy64 (approved):

tap_square/streams.py
asaf-erlich: I added a log line somewhere and then removed it. I don't think it hurts to leave it at the top.

PR #56, opened 2020-08-19
wTesting/data diversity

kspeer825:
# Description of change
Add explicit data manipulation to bookmarks test for `inventories`.
Add cleanup to pagination test for `categories`.
Add emergency cleanup to handle the apparent 10000 record limit for catalog streams.
Add a test to ensure all tap-tester tests are running in circleci.
Add `test_payments.py` which ensures we are catching all possible fields for the `payments` stream (WIP).
# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -25,10 +25,6 @@ jobs:
       #      source /usr/local/share/virtualenvs/tap-square/bin/activate
       #      pylint tests -d 'broad-except,chained-comparison,empty-docstring,fixme,invalid-name,line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,no-else-raise,no-else-return,too-few-public-methods,too-many-arguments,too-many-branches,too-many-lines,too-many-locals,ungrouped-imports,wrong-spelling-in-comment,wrong-spelling-in-docstring,duplicate-code,no-name-in-module,attribute-defined-outside-init'
       - run:
-          name: 'ALl Tests Running'
-          command: |
-            python tests/test_config.py
-      - run:
           name: 'JSON Validator'
           command: |
             source /usr/local/share/virtualenvs/tap-tester/bin/activate
@@ -164,7 +160,13 @@ jobs:
                      --password=$SANDBOX_PASSWORD \
                      --client-id=50 \
                      tests/test_start_date.py
-
+    - run:
+          name: 'All Tests Running'
+          command: |
+            source dev_env.sh
+            source /usr/local/share/virtualenvs/tap-tester/bin/activate
+            PYTHONPATH=$PYTHONPATH:/usr/local/share/virtualenvs/tap-square/lib/python3.5/site-packages/ \
+            python tests/test_config.py
 workflows:
   version: 2
   commit:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,7 +34,8 @@ jobs:
           name: 'Integration Tests Setup'
           command: |
             aws s3 cp s3://com-stitchdata-dev-deployment-assets/environments/tap-tester/sandbox dev_env.sh
-    - run:
+      - run:
+          when: always
           name: 'All Tests Running'
           command: |
             source dev_env.sh
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -34,6 +34,13 @@ jobs:
           name: 'Integration Tests Setup'
           command: |
             aws s3 cp s3://com-stitchdata-dev-deployment-assets/environments/tap-tester/sandbox dev_env.sh
+    - run:
+          name: 'All Tests Running'
+          command: |
+            source dev_env.sh
+            source /usr/local/share/virtualenvs/tap-tester/bin/activate
+            PYTHONPATH=$PYTHONPATH:/usr/local/share/virtualenvs/tap-square/lib/python3.5/site-packages/ \
+            python tests/test_config.py
       - run:
           when: always
           name: 'Testing Discovery'
@@ -160,13 +167,6 @@ jobs:
                      --password=$SANDBOX_PASSWORD \
                      --client-id=50 \
                      tests/test_start_date.py
-    - run:
-          name: 'All Tests Running'
-          command: |
-            source dev_env.sh
-            source /usr/local/share/virtualenvs/tap-tester/bin/activate
-            PYTHONPATH=$PYTHONPATH:/usr/local/share/virtualenvs/tap-square/lib/python3.5/site-packages/ \
-            python tests/test_config.py
 workflows:
   version: 2
   commit:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -168,6 +168,20 @@ jobs:
                      --password=$SANDBOX_PASSWORD \
                      --client-id=50 \
                      tests/test_start_date.py
+      - run:
+          when: always
+          name: 'Testing Payments Stream'
+          command: |
+            source dev_env.sh
+            source /usr/local/share/virtualenvs/tap-tester/bin/activate
+            PYTHONPATH=$PYTHONPATH:/usr/local/share/virtualenvs/tap-square/lib/python3.5/site-packages/ \
+            run-test --tap=tap-square \
+                     --target=target-stitch \
+                     --orchestrator=stitch-orchestrator \
+                     --email=harrison+sandboxtest@stitchdata.com \
+                     --password=$SANDBOX_PASSWORD \
+                     --client-id=50 \
+                     tests/test_payments.py
 workflows:
   version: 2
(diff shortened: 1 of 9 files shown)


.circleci/config.yml
```diff
@@ -146,7 +154,34 @@ jobs:
...
+            source dev_env.sh
+            source /usr/local/share/virtualenvs/tap-tester/bin/activate
+            PYTHONPATH=$PYTHONPATH:/usr/local/share/virtualenvs/tap-square/lib/python3.5/site-packages/ \
+            run-test --tap=tap-square \
+                     --target=target-stitch \
+                     --orchestrator=stitch-orchestrator \
+                     --email=harrison+sandboxtest@stitchdata.com \
+                     --password=$SANDBOX_PASSWORD \
+                     --client-id=50 \
+                     tests/test_start_date.py
```
kspeer825: oops will drop this (duplicate again)

asaf-erlich:
>Add a test to ensure all tap-tester tests are running in circleci.

Nice!

tests/test_config.py
```diff
@@ -0,0 +1,31 @@
+import os
+
+
+cwd = os.getcwd()
+if cwd not in {'root/project', '/opt/code/tap-square'}:
+    print("WARN: This script is meant to run from the top level directory of the tap.")
```
asaf-erlich, line 6: Will these directories work in circle? I assume so but double checking. Also, there's no reason this test shouldn't be able to run from any directory (like ~/git/tap-square). But I guess this is just a print warning and not an assertion... but I'm not sure what value it provides.

tests/test_config.py
```diff
@@ -0,0 +1,31 @@
...
+
+print("Reading in filenames from tests directory.")
+print("Parsing directory for tests.")
```
asaf-erlich, line 9: You print these two, but can you print the files found in a line below?

tests/test_payments.py
```diff
@@ -0,0 +1,131 @@
...
+            if set(record.keys()).difference(fields):
+                print("\n\nNew fields found: update on a {} {}".format(desc, stream))
+                print("Adding untracked fields to tracked set: {}\n\n".format(set(record.keys()).difference(fields)))
+            fields.update(record.keys())
+
+        # Verify all the fields found in the created and updated records are accounted for by the schema
+        schema_keys = set(self.expected_schema_keys(stream))
+        self.assertTrue(fields.issubset(schema_keys),
+                        msg="Fields missing from schema: {}".format(fields.difference(schema_keys)))
+
```
kspeer825, line 128: I want to add a sync and get the sum of keys in replicated records to ensure that no fields were added later due to some workflow. If this is the case, the way the test is currently setup we should catch those fields in a subsequent run of the test_all_fields.py

tests/test_config.py
```diff
@@ -0,0 +1,31 @@
...
+    contents = config.read()
+
+# Check that each file in the directory can be found in a run black in the circle config
+print("Parsing circle config for run blocks.")
+runs = contents.replace(' ', '').replace('\n', '').split('-run:') # separate into run blocks
+matches = {f: False for f in files}
+for m in matches.keys():
+    print("Verifying {} is running in circle.".format(m))
+    if any([m in run for run in runs]):
+        matches[m] = True
```
asaf-erlich, line 28: I think this for loop is a bit complicated. What value is it providing over just asserting all([file in contents for file in files])? Why split it into runs and create a dict of matches[file] to true/false?

asaf-erlich: It's not a huge deal, the test is checking the right thing. I assume you tested it fails and succeeds correctly?

tests/test_pagination.py
```diff
@@ -44,6 +44,16 @@ def testable_streams_static(self):
             'locations',  # This stream does not paginate in the sync (See Above)
         })
 
+    def cleanup_excess(self):
+        cleanup = {'categories': 10000}
+        for stream, limit in cleanup.items():
+            print("Checking if cleanup is required.")
+            all_records = self.client.get_all(stream, start_date=self.STATIC_START_DATE)
```
asaf-erlich: Any reason to use the static start date here and not None or something like that? Does it work if you don't pass a start_time here? I suspect it does but not sure.

tests/test_pagination.py
```diff
@@ -69,6 +79,7 @@ def test_run(self):
         self.assertEqual(set(), self.TESTABLE_STREAMS,
                          msg="Testable streams exist for this category.")
         print("\tThere are no testable streams.")
+        self.cleanup_excess()
```
asaf-erlich: Can you put this in a teardown method? Probably in a https://docs.python.org/3/library/unittest.html#unittest.TestCase.tearDownClass method?

tests/test_config.py
kspeer825: It does matter where it is ran from. It works in those two directories. I think it will fail anywhere else because it will mess up the path. I'm not great at working with paths, not really sure how to make it work from anywhere.

tests/test_payments.py
```diff
@@ -0,0 +1,131 @@
...
+        # Track all fields from the updated records
+        for desc, record in updated_records[stream].items():
+            if set(record.keys()).difference(fields):
+                print("\n\nNew fields found: update on a {} {}".format(desc, stream))
+                print("Adding untracked fields to tracked set: {}\n\n".format(set(record.keys()).difference(fields)))
+            fields.update(record.keys())
+
+        # Verify all the fields found in the created and updated records are accounted for by the schema
+        schema_keys = set(self.expected_schema_keys(stream))
```
asaf-erlich, line 125: So in my testing I found multiple times that it actually failed in the assert dict step because some schema values are not in the top-level keys. Not sure if you want to account for going deeper here?

tests/test_pagination.py
kspeer825: I could use some help with this. I did put it in a teardown originally but it was running multiple times and I couldn't figure out why.

tests/test_config.py
kspeer825: Yeah I think this could use a refactor, it isn't super clear. I was a bit lazy and didn't want to spend too much time parsing the file, I also wasn't sure how to put in an assertion without failing too early. Like I didn't want to fail at the first missing file and skip the rest.

tests/test_pagination.py
kspeer825: I assumed a start date was required. Checking now

kspeer825: The test method for get_all does not work without a start_date param. And I don't think the tap should. Actually, not sure if it matters one way or the other.

tests/test_config.py
kspeer825: Done

.circleci/config.yml
kspeer825: Done

tests/test_config.py
kspeer825: Yes I tested both for circle and local runs.

tests/test_payments.py
kspeer825: I don't think I want to spend the time on this now. But that is good to know, I hadn't considered subfields with this. Harrison pointed out it is redundant to check against the schema files. I mostly like having it in the all fields test because it prints fields that are in the schema but that we don't get data for as a warning. But this can be accounted for by adding a sync and doing the dict equals like you said. I'll do that.

tests/test_pagination.py
asaf-erlich: Ok it's fine, just leave it.

tests/test_config.py
asaf-erlich: It's fine, better than not having a test there.

tests/test_config.py
asaf-erlich: I'm confused why that is the case if you're using cwd as the prefix in the file search... But there is a way in python to expand a path to the full path in case that is necessary. Leave this for now, someone can fix it later when necessary.

tests/test_pagination.py
asaf-erlich: I assume this is because of the issues mentioned earlier regarding the base class not extending testcase. It probably runs every method starting with the "test" prefix (if you just use tearDown). So tearDownClass should only run once, theoretically.

tests/test_payments.py
asaf-erlich: You could argue that's covered by the all fields test, up to you. And yeah, don't need to spend time on it now.

asaf-erlich (approved):

tests/test_pagination.py
kspeer825: Got this working I was using tearDown() before not the class method tearDownClass. I think that was my issue

PR #57, opened 2020-08-19
Testing/fix circle config

kspeer825:
# Description of change
Refactor test_config.py so it is clear and can be run from more directories than just `/tap-square`.
Small fix for teardown (cleanup) method in `pagination_test.py`

# Manual QA steps
 - Tested test_config.py locally in the fail and passing states.
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -24,20 +24,22 @@ cwd = go_to_tests_directory()
 print("Reading in filenames from tests directory.")
 files_in_dir = [name for name in os.listdir(cwd)]
 
-print("Parsing directory for tests.")
+print("Dropping files that are not of the form 'test_<feature>.py'.")
 test_files_in_dir = [fn for fn in files_in_dir if 'test' == fn[:4] and '.py' == fn[-3:]]
-test_files_in_dir.remove('test_client.py')
-test_files_in_dir.remove('test_config.py')
-print("\tFiles found: {}".format(test_files_in_dir))
 
-print("Reading contents of circle config")
+print("Dropping test_client.py from test files.")
+test_files_in_dir.remove('test_client.py')
+
+print("Files found: {}".format(test_files_in_dir))
+
+print("Reading contents of circle config.")
 with open(cwd + "/../.circleci/config.yml", "r") as config:
     contents = config.read()
 
 print("Parsing circle config for run blocks.")
 runs = contents.replace(' ', '').replace('\n', '').split('-run:')
 
-print("Verify all test files are executed in circle")
+print("Verify all test files are executed in circle...")
 file_found = {f: False for f in test_files_in_dir}
 for filename in file_found.keys():
     print("\tVerifying {} is running in circle.".format(filename))
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -18,7 +18,7 @@ def go_to_tests_directory():
 ### TEST
 ##########################################################################
 
-print("Acquiring path to tests direcotry.")
+print("Acquiring path to tests directory.")
 cwd = go_to_tests_directory()
 
 print("Reading in filenames from tests directory.")
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -25,7 +25,7 @@ print("Reading in filenames from tests directory.")
 files_in_dir = [name for name in os.listdir(cwd)]
 
 print("Dropping files that are not of the form 'test_<feature>.py'.")
-test_files_in_dir = [fn for fn in files_in_dir if 'test' == fn[:4] and '.py' == fn[-3:]]
+test_files_in_dir = [fn for fn in files_in_dir if fn.startswith('test_') and fn.endswith('.py')]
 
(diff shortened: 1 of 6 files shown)


tests/test_config.py
```diff
@@ -1,32 +1,50 @@
...
+    'tap-square/tests/',
+    '../tap-square/tests/',
+]
+
+def go_to_tests_directory():
+    for path in potential_paths:
+        if os.path.exists(path):
+            os.chdir(path)
+            return os.getcwd()
+    raise NotImplementedError("This check cannot run from {}".format(os.getcwd()))
```
asaf-erlich, line 15: Again I really think if done correctly this should be able to run from anywhere but I'm ok with leaving this for now.

tests/test_config.py
```diff
@@ -1,32 +1,50 @@
...
-files = [
-    name for name in os.listdir(cwd + '/tests')
-    if 'test' == name[:4] and '.py' == name[-3:] and \
-    name not in ['test_client.py', 'test_config.py']
-]
-print("\tFiles found: {}".format(files))
+files_in_dir = [name for name in os.listdir(cwd)]
+
+print("Dropping files that are not of the form 'test_<feature>.py'.")
+test_files_in_dir = [fn for fn in files_in_dir if 'test' == fn[:4] and '.py' == fn[-3:]]
```
asaf-erlich: You say in the print "test_" but you only check for "test" prefix. Next can you use string startswith and endswith for the strings instead of the `[:4]` and `[-3:]` because I feel they're easier to read?

tests/test_config.py
```diff
@@ -1,32 +1,50 @@
...
-# Verify all files were found
-assert all(matches.values()), "The following tests are not running in circle:\t{}".format([k for k, v in matches.items() if not v])
+runs = contents.replace(' ', '').replace('\n', '').split('-run:')
+
+print("Verify all test files are executed in circle...")
+file_found = {f: False for f in test_files_in_dir}
+for filename in file_found.keys():
+    print("\tVerifying {} is running in circle.".format(filename))
+    if any([filename in run for run in runs]):
+        file_found[filename] = True
```
asaf-erlich: I still feel like you don't need the file_found loop here. What's the point of having the key with false / true values? Why not create an empty set and add to the set if it's fine / start with a full set and remove from the set when it's found?

tests/test_config.py
kspeer825: I agree, I just couldn't figure out how to get that working.

tests/test_config.py
```diff
@@ -1,32 +1,50 @@
...
-    if any([m in run for run in runs]):
-        matches[m] = True
-
-# Verify all files were found
-assert all(matches.values()), "The following tests are not running in circle:\t{}".format([k for k, v in matches.items() if not v])
+runs = contents.replace(' ', '').replace('\n', '').split('-run:')
+
+print("Verify all test files are executed in circle...")
+file_found = {f: False for f in test_files_in_dir}
+for filename in file_found.keys():
```
asaf-erlich: You don't need to do .keys(). If you just do `for filename in file_found` it works, iterating over keys is the default iterator for a dict.

tests/test_config.py
```diff
@@ -1,32 +1,50 @@
...
-    print("Verifying {} is running in circle.".format(m))
-    if any([m in run for run in runs]):
-        matches[m] = True
-
-# Verify all files were found
-assert all(matches.values()), "The following tests are not running in circle:\t{}".format([k for k, v in matches.items() if not v])
+runs = contents.replace(' ', '').replace('\n', '').split('-run:')
+
+print("Verify all test files are executed in circle...")
+file_found = {f: False for f in test_files_in_dir}
```
asaf-erlich: Can you rename this to files_found to indicate it's a collection?

tests/test_config.py
kspeer825: Done

tests/test_config.py
kspeer825: I don't seem much of difference between those approaches but I can do that.

kspeer825: Done

tests/test_config.py
```diff
@@ -1,32 +1,49 @@
...
-assert all(matches.values()), "The following tests are not running in circle:\t{}".format([k for k, v in matches.items() if not v])
+runs = contents.replace(' ', '').replace('\n', '').split('-run:')
+
+print("Verify all test files are executed in circle...")
+tests_not_found = set(test_files_in_dir)
+for filename in test_files_in_dir:
+    print("\tVerifying {} is running in circle.".format(filename))
+    if any([filename in run for run in runs]):
+        tests_not_found.remove(filename)
+assert tests_not_found == set(), "The following tests are not running in circle:\t{}".format(tests_not_found)
```
asaf-erlich, line 48: I would prefer `.isEmpty() vs == set()` preferably. It's not a big deal though

asaf-erlich (approved):
lgtm

PR #60, opened 2020-08-20
Testing/payments all fields

kspeer825:
# Description of change
Merge the creates and updates from  `test_payments.py` into `test_all_fields.py`.
Cleanup some base assertions (reusable vs tap-specific).
Break up `run_initial_sync` into methods for check table/field selection, and sync.
Some general test cleanup.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch


tests/base.py
```diff
@@ -562,32 +568,54 @@ def run_initial_sync(self, environment, data_type, select_all_fields=True): # RE
...
         expected_pks_set = set(expected_pks)
         self.assertEqual(len(expected_pks), len(expected_pks_set), msg="Our expectations contain a duplicate record.")
-        expected_pks_to_record_dict = {tuple(expected_record.get(pk) for pk in primary_keys): expected_record for expected_record in expected_records}
 
         # Verify that all expected records and ONLY the expected records were replicated
-        self.assertEqual(expected_pks_set, actual_pks_set)
+        self.assertEqual(expected_pks_set, sync_pks_set)
 
-        return expected_pks_to_record_dict, actual_pks_to_record_dict
+    def assertParentKeysEqual(self, expected_record, sync_record):
```
asaf-erlich, line 604: More and more I think assertParentKeysEqual should just be assertKeysEqual since "parent" is not adding any information... but it's not a huge deal.

tests/test_all_fields.py
```diff
@@ -20,6 +20,70 @@ def testable_streams_dynamic(self):
     def testable_streams_static(self):
         return self.static_data_streams().difference(self.untestable_streams())
 
+    def ensure_dict_object(self, resp_object):
```
asaf-erlich, line 28: Technically this name could be extended to say ensure_dict_or_list_of_single_dict? Not a huge deal, seems like a useful method.

tests/test_all_fields.py
```diff
@@ -20,6 +20,70 @@ def testable_streams_dynamic(self):
...
+        descriptions = {  # (source type, autocomplete)
+            ("card", False),
+            ("card_on_file", False),
+            ("gift_card", False),
+            ("card", True),
+        }
+        for source_key, autocomplete in descriptions:
+            payment_response = self.client._create_payment(autocomplete=autocomplete, source_key=source_key)
+            payment_object = self.ensure_dict_object(payment_response)
+            payment_records.append({'description': (source_key, autocomplete), 'record': payment_object})
```
asaf-erlich: You're passing this dict around... it really feels like it would be easier to read if you created a class or [namedtuple](https://docs.python.org/3.3/library/collections.html?highlight=namedtuple#collections.namedtuple) . Named tuples in python are the recommended way to create an immutable data class / record / case class / datum (different languages call these different things). Personally I think you would have 3 fields, the source key, autocomplete, and record / payment_record.

This becomes more useful if we were to use type hints. The payment_records list could be of type List[PaymentRecordDetails] or something like that. Otherwise you're kind of just hoping that all the dicts in the list have a key of "description" and "record"... which you don't know

asaf-erlich (commented):
overall looks good, made comments

tests/test_all_fields.py
```diff
@@ -20,6 +26,69 @@ def testable_streams_dynamic(self):
...
+            return resp_object
+        elif isinstance(resp_object, list):
+            self.assertEqual(1, len(resp_object),
+                             msg="Multiple objects were returned, but only 1 was expected")
+            self.assertTrue(isinstance(resp_object[0], dict),
+                            msg="Response object is a list of {} types".format(type(resp_object[0])))
+            return resp_object[0]
+        else:
+            raise RuntimeError("Type {} was unexpected.\nRecord: {} ".format(type(resp_object), resp_object))
+    def create_specific_payments(self):
```
asaf-erlich, line 45: Add a line break above

tests/test_all_fields.py
```diff
@@ -20,6 +26,69 @@ def testable_streams_dynamic(self):
...
+            ("card", True),
+        }
+        for source_key, autocomplete in descriptions:
+            payment_response = self.client._create_payment(autocomplete=autocomplete, source_key=source_key)
+            payment_record = PaymentRecordDetails(source_key, autocomplete, self.ensure_dict_object(payment_response))
+            payment_records.append(payment_record)
+
+        return payment_records
+
+    def update_specific_payments(self, payments_to_update: List[PaymentRecordDetails]):
```
asaf-erlich: Type hints are I think only available in python 3.6 or above and unfortunately I don't know if this tap enforces python >= 3.6. I would actually really like it if it was true, but I think by default we go with python >= 3.5. Check the setup.py. I would recommend upgrading it there since this tap hasn't released yet. But then we need to update circle python env to use that one.

tests/test_all_fields.py
```diff
@@ -1,11 +1,17 @@
...
+from collections import namedtuple
+from typing import List
 
 import tap_tester.runner      as runner
+import tap_tester.connections as connections
 
 from base import TestSquareBase, DataType
 
 
+PaymentRecordDetails = namedtuple('PaymentRecordDetails', 'source_key, autocomplete, record')
```
asaf-erlich, line 11: Yeah this is where I expected you to put it

tests/test_all_fields.py
```diff
@@ -20,6 +26,70 @@ def testable_streams_dynamic(self):
...
+        """Perform specifc updates on specific payment records."""
+        updated_records = []
+        print("Updating payment records by completing, canceling and refunding them.")
+        # Update a completed payment by making a refund (payments must have a status of 'COMPLETED' to process a refund)
+        source_key, autocomplete = ("card", True)
+        description = "refund"
+        payment_to_update = [payment.record for payment in payments_to_update if payment.source_key == source_key and payment.autocomplete == autocomplete][0]
+        _, payment_response = self.client.create_refund(self.START_DATE, payment_to_update)
+        payment_object = self.ensure_dict_object(payment_response)
+        updated_records.append({'description': description, 'record': payment_object})
```
asaf-erlich: I thought you would use the same namedtuple here

tests/test_all_fields.py
```diff
@@ -20,6 +26,70 @@ def testable_streams_dynamic(self):
...
+
+        # Update a payment by completing it
+        source_key, autocomplete = ("card_on_file", False)
+        description = "complete"
+        payment_to_update = [payment.record for payment in payments_to_update if payment.source_key == source_key and payment.autocomplete == autocomplete][0]
+        payment_response = self.client._update_payment(payment_to_update.get('id'), obj=payment_to_update, action=description)
+        payment_object = self.ensure_dict_object(payment_response)
+        updated_records.append({'description': description, 'record': payment_object})
```
asaf-erlich: I thought you would use the same namedtuple here

asaf-erlich (approved):
overall lgtm with minor comments

tests/test_all_fields.py
kspeer825: done

tests/test_all_fields.py
kspeer825: done

tests/test_all_fields.py
kspeer825: done

tests/test_all_fields.py
kspeer825: done

PR #63, opened 2020-08-24
Testing/remove tap dependency

kspeer825:
# Description of change
Remove tap dependencies in square test client.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1,17 +1,16 @@
-import uuid
-import random
-import os
-from datetime import datetime, timedelta, timezone
-import requests
 import backoff
+from datetime import datetime, timedelta, timezone
+import os
+import random
+import requests
+import urllib.parse
+import uuid
 
 import singer
+from square.client import Client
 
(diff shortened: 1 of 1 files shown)


asaf-erlich (approved):
lgtm

asaf-erlich:
As discussed earlier long term we eventually want tap-tester to be a library ideally, but for now this is better than what we have right now

tests/test_client.py
```diff
@@ -858,22 +1209,28 @@ def update(self, stream, obj_id, version, obj=None, start_date=None): # pylint:
         else:
             raise NotImplementedError("{} is not implmented".format(stream))
 
-    def update_item(self, obj_id, version, start_date):
+    def update_item(self, obj, version, start_date):
```
asaf-erlich, line 1212: Not super important but I really don't like naming a variable obj or obj_id. I would prefer specific, in this case, item since we're only updating items. Or record if you're being generic for all.

PR #67, opened 2020-08-27
Yearly pagination for settlements

dmosorast:
# Description of change
Settlements, according to [the API docs](https://developer.squareup.com/reference/square/settlements-api/v1-list-settlements) cannot sync more than a year range at a time.

This PR adds date window pagination if the start_date is set to > 1 year ago.

# Manual QA steps
 - Ran through the case where it paginates, and where it doesn't, and examined the date windows used.
 
# Risks
 - Low, it's still as untested (due to data creation restrictions), but this removes an error condition.
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_square/client.py
+++ b/tap_square/client.py
@@ -21,17 +21,23 @@ def get_batch_token_from_headers(headers):
     else:
         return None
 
+def should_not_retry(ex):
+    """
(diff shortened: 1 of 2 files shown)


asaf-erlich (approved):

asaf-erlich (approved):
lgtm

PR #68, opened 2020-08-27
Testing/start dates

kspeer825:
# Description of change
Added test to check that all streams can sync with the default start date and a start date prior to the default.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_default_start_date.py
+++ b/tests/test_default_start_date.py
@@ -78,7 +78,7 @@ class TestSquareStartDateDefault(TestSquareBaseParent.TestSquareBase):
         streams_without_data = self.untestable_streams()
 
         # Verify we can run a check and sync with the given start date
-        record_count_by_stream =  self.run_standard_sync(self.get_environment(), data_type)
+        record_count_by_stream = self.run_standard_sync(self.get_environment(), data_type)
 
         # Verify we get records for streams with data
         for stream in testable_streams:
--- a/tests/test_default_start_date.py
+++ b/tests/test_default_start_date.py
@@ -83,6 +83,6 @@ class TestSquareStartDateDefault(TestSquareBaseParent.TestSquareBase):
         # Verify we get records for streams with data
         for stream in testable_streams:
             if stream in streams_without_data:
-                self.assertEqual(0, record_count_by_stream.get(stream, 0))
+                self.assertGreaterEqual(record_count_by_stream.get(stream, 0), 0)
             else:
                 self.assertGreater(record_count_by_stream.get(stream), 0)
--- a/tests/test_default_start_date.py
+++ b/tests/test_default_start_date.py
@@ -1,0 +1,88 @@
+import random
+from datetime import datetime as dt
+from datetime import timedelta
+
+import tap_tester.connections as connections
+
+from base import TestSquareBaseParent, DataType
+
+
+class TestSquareStartDateDefault(TestSquareBaseParent.TestSquareBase):
+    """
+    Test we can perform a successful sync for all streams with a start date of 1 year ago and older.
+    This makes up for the time partions missed by `test_start_date.py`.
+    """
+    @staticmethod
+    def name():
+        return "tap_tester_start_date_default"
+
+    def testable_streams_dynamic(self):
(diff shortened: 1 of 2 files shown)


tests/test_default_start_date.py
```diff
@@ -0,0 +1,88 @@
...
+    def default_start_date_test(self, data_type, testable_streams):
+        streams_without_data = self.untestable_streams()
+
+        # Verify we can run a check and sync with the given start date
+        record_count_by_stream = self.run_standard_sync(self.get_environment(), data_type)
+
+        # Verify we get records for streams with data
+        for stream in testable_streams:
+            if stream in streams_without_data:
+                self.assertEqual(0, record_count_by_stream.get(stream, 0))
```
asaf-erlich: Can we do greater than or equal to? No reason to fail if records magically show up...

tests/test_default_start_date.py
```diff
@@ -0,0 +1,88 @@
...
+        self.default_start_date_test(DataType.DYNAMIC, self.testable_streams_dynamic().difference(self.sandbox_streams()))
+        self.default_start_date_test(DataType.STATIC, self.testable_streams_static().difference(self.sandbox_streams()))
+
+        days_prior = random.randint(7, 358) # 1 week to 57 weeks
+
+        # Testing start date prior to default
+        self.START_DATE = dt.strftime(default_start_date - timedelta(days=days_prior), self.START_DATE_FORMAT)
+        self.set_environment(self.SANDBOX)
+        self.default_start_date_test(DataType.DYNAMIC, self.testable_streams_dynamic().difference(self.production_streams()))
+        self.default_start_date_test(DataType.STATIC, self.testable_streams_static().difference(self.production_streams()))
```
asaf-erlich: Can you use intersection with sandbox instead of difference of production?

tests/test_default_start_date.py
```diff
@@ -0,0 +1,88 @@
...
+        self.set_environment(self.PRODUCTION)
+        self.default_start_date_test(DataType.DYNAMIC, self.testable_streams_dynamic().difference(self.sandbox_streams()))
```
asaf-erlich: Similar here, intersection of production, since the environment is production. I find it easier to read.

asaf-erlich (approved):
lgtm with minor comments

tests/test_default_start_date.py
kspeer825: Not sure I agree, if records did show up for those 3 untestable streams I think we'd want to know.

tests/test_default_start_date.py
kspeer825: Done

tests/test_default_start_date.py
kspeer825: Done

tests/test_default_start_date.py
asaf-erlich: Why? Why would we want this test to fail for that reason?

kspeer825: Well generally, I think it is better to be as strict as we can and ease back only when necessary like when we hit instability like we are seeing elsewhere now. I think in this case it would be bad if we randomly get data for those untestable streams. Our expectations would be wrong, we are skipping all the feature tests for these streams due to the fact that we can't get data into the account for them. If something drastic enough changes that data appears, I think that should require someone's attention. Either square is messing up in a very bad way, or our expectations are bad and we shouldn't be skipping all those tests.

asaf-erlich: I don't understand why we want daily builds to fail for this reason. That would mean the test would probably be failing for months until someone investigates. And without context it's hard to determine exactly why it's failing quickly. If there are multiple things that might have started failing between that failure and by the time someone looks at it it becomes harder to distinguish test failure reasons...

kspeer825: I made the change here but I don't totally agree with it. I think I understand your point about the test maintenance issue, but I am not sure if writing assertions that break our expectations is good path to address it.

asaf-erlich: Agree to disagree 🤷

PR #69, opened 2020-08-27
Fix/canary test

dmosorast:
# Description of change
It aopears that our canary cross multiplication of creds + streams didn't match reality, and sandbox creds cannot be used to query for roles, bank_accounts, or settlements.

This PR keeps `employees` in the set of "production streams" in the tests for the purposes of not being able to create data for them, but also removes streams which cannot be queried with sandbox creds (roles, bank_accounts, settlements) from the sandbox discovery mode.

# Manual QA steps
 - Ran through these situations manually and watched the tests pass/fail
 
# Risks
 - Low, mostly testing and UX oriented.
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -196,6 +196,8 @@ class TestSquareBaseParent:
             return {
                 'employees',
                 'roles',
+                'bank_accounts',
+                'settlements',
             }
 
         def sandbox_streams(self):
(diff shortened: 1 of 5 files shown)


asaf-erlich (approved):

tests/base.py
```diff
@@ -184,6 +184,11 @@ def expected_metadata(self):
                 },
             }
 
+            if self.get_environment() == self.SANDBOX:
+                return {k: v for k, v in all_streams.items()
+                        if k not in {'roles', 'bank_accounts', 'settlements'}}
```
asaf-erlich: Can you pull those 3 into a constant somewhere?

PR #70, opened 2020-08-27
Add 401 to non-retryable exception codes

dmosorast:
# Description of change
401 Unauthorized doesn't make sense to be retried, so this PR removes that status code from the client's retry pattern.

# Manual QA steps
 - Ran through and confirmed that it works.
 
# Risks
 - Low, it's retry in safe code. Worst case it still retries.
 
# Rollback steps
 - revert this branch


asaf-erlich (approved):

dmosorast:
Closing, due to branch renaming putting this commit in #69

PR #71, opened 2020-08-28
Testing/fix catalog updates

kspeer825:
# Description of change
Fix catalog updates to use the whole <catalog_stream_name>_data object in the update.
Adjust logic in bookmarks tests, to create a new record with every test run and use that record in the update between syncs.

# Manual QA steps
 - Ran locally against QA tokens
 
# Risks
 - None, test change
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -1344,13 +1344,16 @@ class TestClient():
 
     def update_taxes(self, obj, version):
         obj_id = obj.get('id')
+        all_loc = obj.get('present_at_all_locations')
         LOGGER.info('THIS IS WHAT THE TAX RECORD LOOKS LIKE %s', obj)
+
         tax_data = obj.get('tax_data')
         tax_data['name'] = self.make_id('tax')
 
         body = {'batches': [{'objects': [{'id': obj_id,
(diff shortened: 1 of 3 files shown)


tests/test_client.py
```diff
@@ -1317,29 +1320,38 @@ def update_roles_v1(self, obj):
...
-                                                            'amount_money': {'amount': 34500,
-                                                                             'currency': 'USD'}}}]}],
+                                          'discount_data': discount_data,}]}],
                 'idempotency_key': str(uuid.uuid4())}
         return self.post_catalog(body)
 
-    def update_taxes(self, obj_id, version):
+    def update_taxes(self, obj, version):
+        obj_id = obj.get('id')
+        LOGGER.info('THIS IS WHAT THE TAX RECORD LOOKS LIKE %s', obj)
```
asaf-erlich, line 1347: Do we need to log this?

kspeer825: Nope I can remove, was using for debugging at one point

asaf-erlich (approved):
lgtm - minor comment about removing log info line

PR #78, opened 2020-09-18
retry refund create on version mismatch

kspeer825:
# Description of change
Tests have failed twice due to a version mismatch in the create refund. Address this failure by retrying the create. If this proves to be ineffective we may have to rethink this method. 
# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/test_client.py
+++ b/tests/test_client.py
@@ -773,17 +773,14 @@ class TestClient():
             refunds += created_refund
         return refunds
 
-    @backoff.on_exception(
-        backoff.expo,
-        RuntimeError,
-        max_tries=3,
+
         jitter=backoff.full_jitter,
         on_backoff=log_backoff,
     )
     @backoff.on_exception(
         backoff.expo,
-        requests.exceptions.RequestException,
-        max_time=120, # seconds
+        (requests.exceptions.RequestException, RuntimeError),
+        max_tries=3,
         jitter=backoff.full_jitter,
         on_backoff=log_backoff,
     )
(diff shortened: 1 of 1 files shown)


luandy64 (approved):

tests/test_client.py
```diff
@@ -808,6 +808,10 @@ def create_refund(self, start_date, payment_response=None):
 
         refund = self._client.refunds.refund_payment(body)
         if refund.is_error():
+            if retryable and any([err.get('code') == "VERSION_MISMATCH" for err in refund.errors]):
```
cosimon: Can we do this via backoff instead?

kspeer825: We can, I can try to implement that instead.

cosimon: You still have the recursive retry pattern.

cosimon (changes requested):

tests/test_client.py
```diff
@@ -773,6 +773,13 @@ def create_refunds(self, start_date, num_records, payment_response=None):
             refunds += created_refund
         return refunds
 
+    @backoff.on_exception(
+        backoff.expo,
+        requests.exceptions.RequestException,
```
cosimon: This needs to include the RuntimeError from line 821 or make line 821 raise a RequestException.

cosimon (changes requested):

cosimon (approved):

PR #87, opened 2020-12-09
Fix customers stream

cosimon:
# Description of change
Make the customers stream use date windowing to perform the sync, filter on the start and end on updated_at.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_square/client.py
+++ b/tap_square/client.py
@@ -87,7 +87,7 @@ class SquareClient():
 
         if result.is_error():
             error_message = result.errors if result.errors else result.body
-            if 'Service Unavailable' in error_message or 'upstream connect error or disconnect/reset before headers' in error_message:
+            if 'Service Unavailable' in error_message or 'upstream connect error or disconnect/reset before headers' in error_message or result.status_code == 429:
                 raise RetryableError(error_message)
             else:
                 raise RuntimeError(error_message)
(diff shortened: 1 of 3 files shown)


luandy64 (approved):

tap_square/client.py
```diff
@@ -162,23 +162,20 @@ def get_bank_accounts(self):
...
-                "sort": {
-                    "sort_field": "UPDATED_AT",
-                    "sort_order": "ASC"
-                }
             }
         }
 
-        if bookmarked_cursor:
-            body['cursor'] = bookmarked_cursor
+        if cursor is not None:
```
asaf-erlich, line 177: This is so nitpicky I almost want you to ignore this comment, but even though I know this is more explicit, it's usually more pythonic to just do `if cursor:` unless there's a reason to suspect other non-truthy values could be provided, like false or blank... 

To be more specific after writing the above I looked up where the style suggestion comes from, it's google / pep8:
https://google.github.io/styleguide/pyguide.html?showone=True/False_evaluations#2144-decision

>Use the “implicit” false if possible, e.g., if foo: rather than if foo != []:. There are a few caveats that you should keep in mind though:

>* Always use if foo is None: (or is not None) to check for a None value. E.g., when testing whether a variable or argument that defaults to None was set to some other value. The other value might be a value that’s false in a boolean context!

So the question then becomes is the default above None? You're not writing in a default, but maybe everyone agrees the default behavior should only apply for None? Again, I think I'm overthinking this... so feel free to decide whatever you prefer.

tap_square/sync.py
```diff
@@ -72,8 +81,21 @@ def sync(config, state, catalog):
                 )
                 singer.write_state(state)
 
+            elif tap_stream_id == 'customers':
```
asaf-erlich, line 84: Why did this logic need to be inside the sync function as an if block? Why couldn't you push this down to the stream class sync logic? Would it require too much refactoring to avoid this?

cosimon: We wanted to put it in the class' sync function but putting only this one stream's syncing logic in there would make it an exception to the pattern the rest of the tap uses, so we opted to put it here to keep the tap consistent.

tap_square/client.py
cosimon: On line 85 in sync.py we are explicitly setting the cursor to `None` and here we are checking if it is that value.

tap_square/sync.py
asaf-erlich: I disagree with this decision. The whole point of the tap class's sync function is they are different classes, so overriding it there is the right way to do it. Other sync functions (for example, for incremental streams) are different. There's not just a single sync inside it...

And why did you merge without discussing this further?

asaf-erlich: Please ignore the above comment after we zoom synced about this.

PR #91, opened 2020-12-21
Refactor sync code into individual stream classes

asaf-erlich:
# Description of change
Major refactor to clean up the sync.py code and push the differences into the individualized stream classes
- only minor functional change ( get_orders query) but otherwise just refactoring
- Functional changes coming in subsequent PR

# Manual QA steps
 - Sync canary test passes locally
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tap_square/client.py
+++ b/tap_square/client.py
@@ -183,27 +183,25 @@ class SquareClient():
             body,
             'customers')
 
-    def get_orders(self, location_ids, start_time, bookmarked_cursor):
-        if bookmarked_cursor:
-            body = {
-                "cursor": bookmarked_cursor,
-            }
-        else:
-            body = {
-                "query": {
-                    "filter": {
-                        "date_time_filter": {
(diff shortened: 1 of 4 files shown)


KAllan357 (approved):

dmosorast (approved):
I think I understand most of the changes, didn't use a fine-toothed comb, but as long as the tests are passing and any streams that aren't under test are manually tested, I'm good.

asaf-erlich:
>I think I understand most of the changes, didn't use a fine-toothed comb, but as long as the tests are passing and any streams that aren't under test are manually tested, I'm good.

Let's chat about this at standup tomorrow. The only streams not under test require a credit card, so even though we have an account we can't add data to it. When we shipped square we said we would find clients to make sure this works. Do we need to lookup some clients we would ask to clone connections for? Or what's the best path forward?

PR #92, opened 2020-12-22
Remove bookmark cursor from incremental streams and test bookmark cursor for every stream using it

asaf-erlich:
# Description of change
-Remove bookmark cursor from incremental streams and test bookmark cursor for every stream using it
-If we get any exception (after retrying) while querying for a stream we clear the pagination bookmark and re-raise the exception (so we still see a hard failure on that tap run and it should be able to recover next time). This is to avoid cursor timing out which is another bug that at least one of the support tickets had. There was no great way to distinguish a cursor timing out or a bad cursor value or other bad requests. It lets the stream be interruptible, but avoids letting customers get into a non-recoverable state.

Streams which we removed the bookmark cursor from because they bookmark already with a replication key:
1. customers
1. orders
1. All catalog streams: items, categories, discounts, taxes, modifier lists 

Streams we added back / fixed the cursor functionality to that were missing it / not using it correctly:
1. shifts
1. employees
1. roles

Explanation of new test_bookmarks_cursor test:

1. In the test client call the same exact query that the tap uses, grab first page of values and the cursor value, don’t call the query again with that cursor
1. Shove that cursor value into the state / bookmarked cursor value before starting the tap.
1. Start the tap, expecting it to use that bookmarked cursor value.
1. Compare the pks of records from the tap against all the records (using get_all which iterates through all pages) and subtracting the first page value pks
 
# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,7 @@ jobs:
           name: 'pylint tap'
           command: |
             source /usr/local/share/virtualenvs/tap-square/bin/activate
-            pylint tap_square -d 'broad-except,chained-comparison,empty-docstring,fixme,invalid-name,line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,no-else-raise,no-else-return,too-few-public-methods,too-many-arguments,too-many-branches,too-many-lines,too-many-locals,ungrouped-imports,wrong-spelling-in-comment,wrong-spelling-in-docstring,too-many-public-methods,missing-docstring'
+            pylint tap_square -d 'broad-except,chained-comparison,empty-docstring,fixme,invalid-name,line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,no-else-raise,no-else-return,too-few-public-methods,too-many-arguments,too-many-branches,too-many-lines,too-many-locals,ungrouped-imports,wrong-spelling-in-comment,wrong-spelling-in-docstring,too-many-public-methods,missing-docstring,arguments-differ'
       - run:
           name: 'pylint tests'
           command: |
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -13,7 +13,7 @@ jobs:
       - checkout
       - add_ssh_keys
       - run:
-          name: 'setup virtual env'
+          name: 'Setup virtual env'
           command: |
             python3 -m venv /usr/local/share/virtualenvs/tap-square
             source /usr/local/share/virtualenvs/tap-square/bin/activate
@@ -31,7 +31,7 @@ jobs:
             source /usr/local/share/virtualenvs/tap-square/bin/activate
             pylint tests -d 'broad-except,chained-comparison,empty-docstring,fixme,invalid-name,line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,no-else-raise,no-else-return,too-few-public-methods,too-many-arguments,too-many-branches,too-many-lines,too-many-locals,ungrouped-imports,wrong-spelling-in-comment,wrong-spelling-in-docstring,duplicate-code,no-name-in-module,attribute-defined-outside-init,too-many-statements,cell-var-from-loop,too-many-public-methods,missing-docstring'
       - run:
-          name: 'json validator'
+          name: 'JSON Validator'
           command: |
             source /usr/local/share/virtualenvs/tap-tester/bin/activate
             stitch-validate-json /usr/local/share/virtualenvs/tap-square/lib/python3.5/site-packages/tap_square/schemas/*.json
@@ -43,11 +43,11 @@ jobs:
       - checkout
       - add_ssh_keys
       - run:
-          name: 'integration tests setup'
+          name: 'Integration Tests Setup'
           command: |
             aws s3 cp s3://com-stitchdata-dev-deployment-assets/environments/tap-tester/sandbox dev_env.sh
       - run:
-          name: 'all tests running'
+          name: 'All Tests Running'
           command: |
             source dev_env.sh
             source /usr/local/share/virtualenvs/tap-tester/bin/activate
@@ -60,7 +60,7 @@ jobs:
       - checkout
       - add_ssh_keys
       - run:
-          name: 'setup virtual env'
+          name: 'Setup virtual env'
           command: |
             python3 -m venv /usr/local/share/virtualenvs/tap-square
             source /usr/local/share/virtualenvs/tap-square/bin/activate
@@ -68,11 +68,11 @@ jobs:
             pip install --upgrade setuptools
             pip install .[dev]
       - run:
-          name: 'integration tests setup'
+          name: 'Integration Tests Setup'
           command: |
             aws s3 cp s3://com-stitchdata-dev-deployment-assets/environments/tap-tester/sandbox dev_env.sh
       - run:
-          name: 'test discovery'
+          name: 'Test Discovery'
           command: |
             source dev_env.sh
             source /usr/local/share/virtualenvs/tap-tester/bin/activate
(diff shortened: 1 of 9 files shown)


tap_square/streams.py
```diff
@@ -253,17 +260,19 @@ class Inventories(FullTableStream):
     replication_key = None
 
     def get_pages(self, bookmarked_cursor, start_time):
-        for page, cursor in self.client.get_inventories(start_time, bookmarked_cursor):
-            yield page, cursor
+        yield from self.client.get_inventories(start_time, bookmarked_cursor)
 
 
-class Shifts(Stream):
+class Shifts(FullTableStream):
```
kspeer825, line 266: I think we should update the metadata in tests/base.py to reflect the change from incremental to full table. I think this should cause the bookmarks to fail if it isn't correct.

tests/test_bookmarks_cursor.py
```diff
@@ -0,0 +1,128 @@
...
+    @classmethod
+    def tearDownClass(cls):
+        print("\n\nTEST TEARDOWN\n\n")
+
+    def test_run(self):
+        """Instantiate start date according to the desired data set and run the test"""
+
+        self.START_DATE = self.get_properties().get('start_date')
+
+        print("\n\nTESTING WITH DYNAMIC DATA IN SQUARE_ENVIRONMENT: {}".format(os.getenv('TAP_SQUARE_ENVIRONMENT')))
```
kspeer825: This can be dropped I think it's a duplicate.

tests/test_bookmarks_cursor.py
```diff
@@ -0,0 +1,128 @@
...
+        self.bookmarks_test(self.testable_streams_dynamic().intersection(self.sandbox_streams()))
+
+        self.set_environment(self.PRODUCTION)
+        production_testable_streams = self.testable_streams_dynamic().intersection(self.production_streams())
+        if production_testable_streams:
```
kspeer825, line 48: I don't like this pattern of setting expectations  and running a test using conditional logic. I think we should know if that intersection returns a set and run the test accordingly.

tests/test_client.py
```diff
@@ -514,6 +514,29 @@ def get_all(self, stream, start_date): # pylint: disable=too-many-return-stateme
         else:
             raise NotImplementedError("Not implemented for stream {}".format(stream))
 
+    def get_first_page_and_cursor(self, stream, start_date):
```
kspeer825, line 520: Nice, I like this method.

kspeer825 (changes requested):

tap_square/streams.py
asaf-erlich: It definitely didn't change to be full table. It might have just made sense to reuse the code in the base FullTableStream class. But I'll rename it if that's the case.

tests/test_bookmarks_cursor.py
asaf-erlich: Sounds good, will fix

tests/test_bookmarks_cursor.py
asaf-erlich: I added this because sometimes I want to only run the test for one stream, but if I do that both tests still run which is weird.

KAllan357 (approved):

tap_square/streams.py
asaf-erlich: Renamed to BookmarkCursorStream

tests/test_bookmarks_cursor.py
asaf-erlich: Dropped the print

tests/test_bookmarks_cursor.py
asaf-erlich: Going to leave this pattern in, let me know if it's a problem

tap_square/streams.py
asaf-erlich: Now that I'm going through the code this name doesn't make sense either, not all of the classes that implement this actually use the bookmark cursor, I'm going to revert it to full table stream, shifts is the only exception, it's not really a full table stream, though in the api code we treat it just like a full table stream, so not sure... maybe call it ClearBookmarkCursorOnExceptionStream, but that's way too implementation specific.

I'm going to revert it to FullTableStream for now, again, if you look in shifts it says the replication method is INCREMENTAL, not FULL, so I think that's more definitive

tests/base.py
```diff
@@ -625,6 +649,9 @@ def assertPKsEqual(self, stream, expected_records, sync_records):
             # Verify sync pks have all expected records pks in it
             self.assertTrue(sync_pks_set.issuperset(expected_pks_set))
 
+            if assert_pk_count_same:
```
asaf-erlich, line 652: @kspeer825 this is the flag I added which make the test work as expected. Refunds and payments both ignore the cursor and return all pages if an invalid cursor / cursor that doesn't match the previous query is sent.

tests/test_bookmarks_cursor.py
kspeer825: If it makes it easier to work with then I think it's fine.

tests/test_client.py
```diff
@@ -934,10 +967,10 @@ def create_modifier_list(self, num_records):
 
     def _create_item(self, start_date, num_records=1):
         mod_lists = self.get_all('modifier_lists', start_date)
-        if mod_lists:
-            mod_list_id = random.choice(mod_lists).get('id')
-        else:
-            mod_list_id = self.create_modifier_list(num_records).body.get('objects').get('id')
+        if not mod_lists:
+            mod_lists = self.create_modifier_list(1).body.get('objects')
```
kspeer825, line 971: Why was this hard coded to 1 rather than using `num_records`? It might be fine, I just can't remember if we were intentionally trying to get 1 to 1 relationship between streams or if it doesn't matter.

tests/base.py
```diff
@@ -571,15 +591,16 @@ def perform_and_verify_table_and_field_selection(self, conn_id, found_catalogs,
...
             Clear the connections state in menagerie and Run a Sync.
             Verify the exit code following the sync.
 
             Return the connection id and record count by stream
             """
-            #clear state
-            menagerie.set_state(conn_id, {})
+            if clear_state:
+                #clear state
+                menagerie.set_state(conn_id, {})
```
kspeer825, line 603: good catch

kspeer825 (approved):
Left 1 comment, but if tests are passing I think these are good changes.

tests/test_client.py
asaf-erlich: If you look at a few lines below you see that we always grab a random id from the list, so we only actually need 1 mod list to create N items. When create modifier list is called not in this context it will create however many are required.

asaf-erlich: Also `self.create_modifier_list(num_records).body.get('objects').get('id')` was failing because `self.create_modifier_list(num_records).body.get('objects')` returns a list, not the first object's dictionary value. I think a previous refactor left this bad code in there from a past time.

.circleci/config.yml
```diff
@@ -31,150 +31,229 @@ jobs:
             source /usr/local/share/virtualenvs/tap-square/bin/activate
             pylint tests -d 'broad-except,chained-comparison,empty-docstring,fixme,invalid-name,line-too-long,missing-class-docstring,missing-function-docstring,missing-module-docstring,no-else-raise,no-else-return,too-few-public-methods,too-many-arguments,too-many-branches,too-many-lines,too-many-locals,ungrouped-imports,wrong-spelling-in-comment,wrong-spelling-in-docstring,duplicate-code,no-name-in-module,attribute-defined-outside-init,too-many-statements,cell-var-from-loop,too-many-public-methods,missing-docstring'
       - run:
-          name: 'JSON Validator'
+          name: 'json validator'
```
kspeer825: What's up with all the lower case?

asaf-erlich: That was a mistake, I selected all text and hit the button to lowercase everything. if you look at the diff now those lowercases should all be gone.

PR #102, opened 2021-03-15
run pylint with tap-tester not tap venv

kspeer825:
# Description of change
For pylint ran against the tests, we should use the tap-tester venv.

Also new fields were added by square to the payments stream.  A backlog ticket was created and workarounds were implemented in the test for those fields.

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch

### DIFF
--- a/tests/base.py
+++ b/tests/base.py
@@ -693,7 +693,7 @@ class TestSquareBaseParent:
 
         def assertDictEqualWithOffKeys(self, expected_record, sync_record, off_keys=frozenset()):
             """
-            This method is currenlty being used to enable us to compare records for streams which 
+            This method is currenlty being used to enable us to compare records for streams which
             have known discrepancies.
 
             Currently those known discrepancies include:
--- a/tests/base.py
+++ b/tests/base.py
@@ -692,15 +692,38 @@ class TestSquareBaseParent:
                 self.assertDictEqual(expected_record, sync_record)
 
         def assertDictEqualWithOffKeys(self, expected_record, sync_record, off_keys=frozenset()):
-            # self.assertParentKeysEqual(expected_record, sync_record)
+            """
+            This method is currenlty being used to enable us to compare records for streams which 
+            have known discrepancies.
+
+            Currently those known discrepancies include:
+             -  https://stitchdata.atlassian.net/browse/SRCE-4975
+             -  https://stitchdata.atlassian.net/browse/SRCE-5143
+            """
+
             expected_record_copy = deepcopy(expected_record)
             sync_record_copy = deepcopy(sync_record)
 
-            # Square api workflow updates these values so they're a few seconds different between
-            # the time the record is created and the tap syncs, but other fields are the same
             for off_key in off_keys:
                 if sync_record_copy.get(off_key):
                     sync_record_copy.pop(off_key)
                 if expected_record_copy.get(off_key):
                     expected_record_copy.pop(off_key)
             self.assertDictEqual(expected_record_copy, sync_record_copy)
+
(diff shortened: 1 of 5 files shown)


tests/base.py
```diff
@@ -692,13 +692,15 @@ def assertRecordsEqual(self, stream, expected_record, sync_record):
                 self.assertDictEqual(expected_record, sync_record)
 
         def assertDictEqualWithOffKeys(self, expected_record, sync_record, off_keys=frozenset()):
-            self.assertParentKeysEqual(expected_record, sync_record)
```
asaf-erlich, line 695: Why would this line be commented? It's just asserting the top-level keys are identical?

tests/test_bookmarks.py
```diff
@@ -485,11 +485,20 @@ def bookmarks_test(self, testable_streams):
...
+
+                        # BUG_2 | https://stitchdata.atlassian.net/browse/SRCE-5143
+                        MISSING_FROM_SCHEMA = {'payments': {'capabilities', 'version_token', 'approved_money'}}
+
                         # Test Workaround Start ##############################
-                        # BUG | https://stitchdata.atlassian.net/browse/SRCE-4975
+                        # BUG_1 | https://stitchdata.atlassian.net/browse/SRCE-4975
                         if stream == 'payments':
+
+                            PARENT_FIELD_MISSING_SUBFIELDS = {'card_details'} # BUG_1
```
asaf-erlich: Capital letters like this for python should be used for constants, but this is just a variable. Can you pull it out into a constant outside the method instead of updating it inside this if statement?

asaf-erlich (changes requested):
Made a few minor comments

asaf-erlich (approved):
On second thought all comments are pretty minor, approved with your discretion though I think the pulling out the constant or making it lowercase is somewhat more important for readability

tests/base.py
kspeer825: Anything that would fail here, we would catch in the assertions in `test_all_fields.py`. Since there are new fields for `payments` now I am already implementing a workaround in the tests for it. If we wanted to keep it we'd need two of these OffKeys assertion methods; one method that checks parent keys match, and another that does not. But since it is covered elsewhere I dropped it.

tests/test_bookmarks.py
kspeer825: Addressed this in most recent commit

tests/base.py
kspeer825: Addressed this in most recent commit.

tests/test_bookmarks.py
```diff
@@ -485,11 +499,18 @@ def bookmarks_test(self, testable_streams):
...
+
                         # Test Workaround Start ##############################
-                        # BUG | https://stitchdata.atlassian.net/browse/SRCE-4975
                         if stream == 'payments':
+
+                            off_keys = MISSING_FROM_SCHEMA[stream] # BUG_2
+                            self.assertParentKeysEqualWithOffKeys(
+                                updated_record, sync_record, off_keys
+                            )
+                            off_keys = PARENT_FIELD_MISSING_SUBFIELDS[stream] | MISSING_FROM_SCHEMA[stream] # BUG_1 | # BUG_2
```
asaf-erlich, line 510: I prefer explicitly calling union, not everyone knows `|` means union for sets. But it's not a huge deal.

asaf-erlich (approved):

tests/test_bookmarks.py
kspeer825: Oh good point. It would be more clear that way. If this PR wasn't blocking Leslie and Zach I would fix it now. I will make a note in the card I spun off to improve readability of these tests.

PR #109, opened 2023-06-20
TDL-23235 add payouts stream

rdeshmukh15:
# Description of change
- Removed settlements stream as it is deprecated and replaced it with the payouts stream ([TDL-23235](https://jira.talendforge.org/browse/TDL-23235))
- API version upgrade

Currently, integration test cases are failing with error access token creation quota limit exceeded which will be fixed in different PR

# Manual QA steps
 - 
 
# Risks
 - 
 
# Rollback steps
 - revert this branch


tap_square/discover.py
```diff
@@ -8,7 +8,7 @@ def get_abs_path(path):
     return os.path.join(os.path.dirname(os.path.realpath(__file__)), path)
 
 # NB: These streams cannot be queried using Sandbox OAuth credentials
-PRODUCTION_ONLY_STREAMS = {'roles', 'bank_accounts', 'settlements'}
+PRODUCTION_ONLY_STREAMS = {'roles', 'bank_accounts'}
```
dsprayberry: should payouts be added here as well or are we able to query the new object from sandbox?

dsprayberry (changes requested):
It looks like settlements is referenced in the test_client file as well which we should probably clean up as part of this: https://github.com/singer-io/tap-square/blob/8fa008a532348b10bca27b9fdd7ce2abfd2613be/tests/test_client.py#L433

tap_square/discover.py
rdeshmukh15: I tried querying the payouts using sandbox creds where I'm getting the empty response {}. So, adding the payouts in prod only streams.

bryantgray (approved):

